PR #26, opened 2020-08-23
[Docs] Add cc by 4 license and badge

### LINKED ISSUE #20: [Discussion] We need to determine the license we use
At start, I commit Apache 2.0 License for the project, I think it is OK for the code part which is the hypertrons workflow components and post processor for SQL components.

But we should also determine what License we use for the docs part which is the report, I think we should protect the report quite carefully, maybe one of CC-BY 4.0 can be considered.

WDYT?

frank-zsy:

Close #20 Add CC-BY-4.0 license and license badges in Readme.

### DIFF
--- a/LICENSE-CC-BY
+++ b/LICENSE-CC-BY
@@ -1,0 +1,395 @@
+Attribution 4.0 International
+
+=======================================================================
+
+Creative Commons Corporation ("Creative Commons") is not a law firm and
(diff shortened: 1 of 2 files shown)


frank-zsy:
@will-ww could you review and approve this PR?

will-ww:
/approve

Great~

PR #31, opened 2020-08-24
[Refactor] Change number to thousand format.

frank-zsy:

In the original version, all numbers of developers and repos are pure number format which is really hard to read, so I change the output format to add comma for every 3 digits so it gonna be better for reading. Like 1234567 -> 1,234,567.


frank-zsy:
@heming6666 Could you review and approve this PR?

heming6666 (approved):
LGTM.

/approve

PR #32, opened 2020-08-24
docs(documentations):add a draft for data description#23

tianyichow:
This is a draft for the data description. We can discuss what is the best way to show the data details.


frank-zsy:
- If you want to upload an image, you may need to create a new folder named `pic` or `img`, because `diagram` folder is for diagrams written in `plantuml`
- I don't think pictures of table is not good enough for the doc, first it is really slow to load a picture, second it is hard to collaborate, we may need to extract the schema to text version and we need to describe every column to make sure developers understand them.

tianyichow:
Yes, I agree with you. It's a better way to show the data table details. There are still some problems that I concern. 
* Examples are also important. What sort of examples we should offer in order to  make newcomers understand this quickly.
* I think we need a Chinese virsion of this data description.

frank-zsy:
Yes, all docs should be in both English and Chinese version. And actually under `sqls` folder, there will be lots of example sqls, I don't think it's a problem.

tianyichow:
I find that inserting a large table into the Markdown file will make it difficult to read. So I propose to present the database description in a csv file.  There is a copy on [Google Doc](https://docs.google.com/spreadsheets/d/1r8731Voz70iavZPSRMGNMsQopYV1FgChZjfzyDfFATo/edit?usp=sharing) to help us to collaborate on data descriptions.

frank-zsy:
Maybe we can split the columns by event types, you can refer to the `parser` of `clickhosue-importer` and see the columns belong to different event types. And it will be better way to organize the docs.

tianyichow:
Most of the data descriptions have been finished. There are some rows that I don't know what should be written.

frank-zsy:
I will update the doc.

frank-zsy:
All absent description added to the sheet. @tianyichow 

Actually I found that you can find most descriptions in the GitHub official docs here: https://developer.github.com/v3/activity/event_types/

tianyichow:
Ok, I will check that.

frank-zsy:
You should not merge the PR yourself, you need another committer to approve your PR.

will-ww:
Can we roll back and try it again ?  :-)  @tianyichow

tianyichow:
Does Pull Request support  roll back for 'merge' action? @frank-zsy

frank-zsy:
> Does Pull Request support roll back for 'merge' action? @frank-zsy

Actually yes, but I think this one is OK and this will be the last one that can be merged by developers because I changed the base permission of org members to `read` so no one has write permission now.

PR #35, opened 2020-08-25
[Feature] Add pr uml render component

### LINKED ISSUE #34: [Feature] UML renderer for PR should be in work flow
We have a lot of diagrams written in UML format and we can render them with Hypertrons too. But if the UML file is edited from a PR, we can not easily check whether it is as expected or not.

For UML files in PR, maybe we can provide a component to render them in PR and comment back to check.

frank-zsy:

Close #34 , I implement the `pr_uml_renderer` component in Hypertrons because I think it could be a very common ability. See https://github.com/X-lab2017/github-analysis-report-2020/pull/33#issuecomment-679890618 for what this component can do.

Others:

- Add `auto_update_contribution` component in config
- For `pull/sql` label, only need to check `created` and `synchronize` actions for `PullRequestEvent`, no need for all kinds of actions.

### DIFF
--- a/.github/hypertrons-components/sql_label/index.lua
+++ b/.github/hypertrons-components/sql_label/index.lua
@@ -15,21 +15,23 @@
 -- Pull label sql
 on('PullRequestEvent', function (e)
   wrap(function()
-    local prNumer = e.number;
-    local files = listPullRequestFiles(e.number)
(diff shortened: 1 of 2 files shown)


heming6666 (approved):
LGTM

PR #48, opened 2020-08-29
[Docs] Change slack join link to badge

frank-zsy:

Change join `Slack` link from text to a badge to better presentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # GitHub Analysis Report 2020
 
-![apache2](https://img.shields.io/badge/license-Apache%202-blue) ![ccby4](https://img.shields.io/badge/license-CC%20BY%204.0-blue)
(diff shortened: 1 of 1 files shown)


frank-zsy:
Also add `Apache license` and `CC-BY-4.0 license` file link on the badge.

sunshinemingo:
I think this is a good idea. But if we change Slack join link to badge, I wonder whether to keep the text about joining Slack.

frank-zsy:
> I think this is a good idea. But if we change Slack join link to badge, I wonder whether to keep the text about joining Slack.

I think we don't need the text so I removed the text part as well, WDYT?

sunshinemingo:
Maybe there is such a situation. Someone may ignore the slack badge when browsing readme file, so that he can not find a way to communicate with us. Maybe we can add a guidance like this in readme: if you are interested in our project, you can join us through slack badge.

frank-zsy:
That is right, I updated the PR.

PR #59, opened 2020-09-02
[Refactor] Contribution sort alphabetically if score is equal

### LINKED ISSUE #58: [Feature] A stable sort algorithm for developers needed
I think we need to add a stable sort algorithm for developers if the score is same to avoid massive move of followers everyday. Maybe the socre can be sorted alphabetically by name.

frank-zsy:

When generate contribution doc, if `score` is equal then sort by `login` so it should be a stable sort.

Close #58


heming6666 (approved):
/approve

PR #76, opened 2020-09-12
[Refactor] change base url

frank-zsy:
First the robot is deployed on my own laptop which is not quite easy to maintain. Now the robot is on cloud and only data service runs on my laptop.

The new base URL is `http://gar2020.opensource-service.cn` so need to change all UML render base URL to make sure they works.


heming6666 (approved):
LGTM.

/approve

PR #89, opened 2020-09-16
[Feature] Add working hour sql component

frank-zsy:
Signed-off-by: Frankzhaopku <syzhao1988@126.com>

Close #83 

Add working hour stats, use `svg` to implement the image

PR #96, opened 2020-09-17
[Fix] Fix few things about working hour

frank-zsy:

Few fixes about working hour, relate to #89 

- Add global config to `runJsCode` in update report
- Fix URL param error
- Add to website in English version because dynamic `svg` file can not be rendered by GitHub markdown renderer.
- Sometimes there are no records for certain `dayOfWeek` and `hour`, so need to return `dayOfWeek` and `hour` and check later for zero values.

### DIFF
--- a/.github/hypertrons-components/auto_update_report/index.lua
+++ b/.github/hypertrons-components/auto_update_report/index.lua
@@ -43,7 +43,7 @@ sched(compConfig.schedName, compConfig.sched, function ()
             ['query'] = sql
           }
         })
-        local renderText = runJsCode(postProcessor, string2table(requestRes).data)
+        local renderText = runJsCode(postProcessor, string2table(requestRes).data, compConfig.defaultRenderParams)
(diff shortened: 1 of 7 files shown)


frank-zsy:
Do not know why online behavior is not same with my local environment.

frank-zsy:
On my laptop the result is as expected, but the CI online give me another result. When `hour` is 0, the result is always `0`. Can figure out what is going on here, can anyone help me here?

frank-zsy:
The bug is caused by Hypertrons Lua Vm which has a bug to convert zero value from Lua to JavaScript, already fixed.

/sql-verified

will-ww:
> The bug is caused by Hypertrons Lua Vm which has a bug to convert zero value from Lua to JavaScript, already fixed.
> 
> /sql-verified

Many thanks, really great to fix this bug.

frank-zsy:
@sunshinemingo could you review this PR please.

sunshinemingo:
Sorry for delay reply.

PR #195, opened 2020-11-21
[Docs] update slack url

sunshinemingo:
The `Slack` url has expired, we need to update it.


frank-zsy:
/approve
Thx for update

PR #248, opened 2020-12-14
Update REPORT_TEMPLATE.md

16647615268:
English replace Englist


16647615268:
English replace Englist

Issue #206, opened 2020-11-27
[Feature] Chinese projects analysis should be added into the report.


frank-zsy:
As in GitHub report 2019 we also need to add Chinese projects analysis. In GitHub report 2019 we use a text file to indicate all main Chinese projects on GitHub to generate the report, we may use the same way to collect all Chinese projects.

But we may need to use repo_id rather than repo_name instead in case name change during 2020, and organization should be supported too.

PR #251, opened 2020-12-15
add Chinese info

wengzhenjie:
#206


wengzhenjie:
I get some information about Chinese developers and organizations by processing the given text. I also get some Chinese full repo name from text. Then I get repo_id and actor_id by using GitHub API. And for developers and organizations, I get all repos under them. So I add more repos.

frank-zsy:
There are several tricks about this component.

- We don't really need to list all repo id of Chinese projects, we can use organization id because there is `org_id` in database which can be used. So the manifest can be quite shorter and also we can cover new projects from the org.
- We need to find a way to inject repo and org id into SQL rather than filter in `post-processor` which will cause a quite huge network load.

sqls/activity-repo-top-Chinese/sql
```diff
@@ -0,0 +1,35 @@
...
+            ON icc.repo_id=rcc.repo_id AND icc.actor_id=rcc.actor_id
+            LEFT JOIN
+            (SELECT repo_id, issue_author_id AS actor_id, COUNT(*) as count,SUM(pull_commits) AS commits,SUM(pull_additions) AS additions,SUM(pull_deletions) AS deletions FROM {{table}} WHERE type='PullRequestEvent' AND action='closed' AND pull_merged=1 GROUP BY repo_id, actor_id HAVING repo_id IN {{[...repos.values()]}}) AS mpc
+            ON icc.repo_id=mpc.repo_id AND icc.actor_id=mpc.actor_id
+        ) AS contribute_list
+        GROUP BY repo_id
+        ORDER BY repo_activity DESC
+        LIMIT {{topN}}) activity_list
+        INNER JOIN
+        (SELECT repo_id,anyHeavy(repo_name) AS repo_name,anyHeavy(repo_language) AS repo_language FROM
```
frank-zsy, line 32: Language stats should be like global SQL

frank-zsy: I didn't realize that in the global rank list SQL, the language query still is another join table. Actually we can use `PullRequestEvent` to fetch the language data because language field only exists in `PullRequestEvent`.

wengzhenjie: Many PRs language is null. If I use same conditions to filter PRs, the number of PRs is not right.

wengzhenjie:
> There are several tricks about this component.
> 
> * We don't really need to list all repo id of Chinese projects, we can use organization id because there is `org_id` in database which can be used. So the manifest can be quite shorter and also we can cover new projects from the org.
> * We need to find a way to inject repo and org id into SQL rather than filter in `post-processor` which will cause a quite huge network load.

OK. I will use org_id to filter repos. However, according to my current stats, the number of actors is very small. Can we get more actors info by other ways?

frank-zsy:
I am not sure what do you mean. Do you mean you want more info than actor info in log data? Like?

wengzhenjie:
> I am not sure what do you mean. Do you mean you want more info than actor info in log data? Like?

Yeah. I only get 60 actors. But we show top 20 active developers. The data looks like small.

frank-zsy:
What do you mean 60 actors? For one repo or what? The global active developer accounts are more than 13 million.

wengzhenjie:
> What do you mean 60 actors? For one repo or what? The global active developer accounts are more than 13 million.

I get actor’s info from text. In the text, the number of Chinese actors is small.

sqls/activity-repo-top-Chinese/sql
frank-zsy: You can use `max` instead of `anyHeavy` to filter the empty values since language is a quite stable column.

frank-zsy:
Oh, I see, actually we can not find out all developer accounts from China, the text accounts mean that all repos under the accounts should be considered as Chinese repos. But we do not make report for Chinese developers because we can not figure out if the contributors account are Chinese or not.

sqls/activity-repo-top-Chinese/sql
wengzhenjie: Can I use max function in string field? Anyway, one repo may have some languages. So I use anyHeavy function to get more frequent language in repo.

wengzhenjie:
> Oh, I see, actually we can not find out all developer accounts from China, the text accounts mean that all repos under the accounts should be considered as Chinese repos. But we do not make report for Chinese developers because we can not figure out if the contributors' accounts are Chinese or not.

Shall we show top Chinese active developers in report?

sqls/activity-repo-top-Chinese/sql
frank-zsy: Yes, you can use `max` on string field. Actually in the log data, only dominate language is included and for most repos(especially most active repos), dominate language does not change. So maybe you can use `max` in first level aggregation and `anyHeavy` for second level aggregation if you concern about language change.

frank-zsy:
> Shall we show top Chinese active developers in report?

I don't think it is necessary and we did not do that in 2019 report too.

sqls/activity-repo-top-Chinese/sql
wengzhenjie: Most PRs language is null. If I don’t judge null condition, the anyHeavy function will return null.

wengzhenjie:
> > Shall we show top Chinese active developers in report?
> 
> I don't think it is necessary and we did not do that in 2019 report too.

OK.

sqls/activity-repo-top-Chinese/sql
wengzhenjie: I agree that for most situations, using max function is enough. I will fix it.

frank-zsy: I think it's fine to collect repo name and language in this way, we need to figure out why so many PR events are without language field.

frank-zsy:
I assume you may need a `pre-processor` to construct your SQL config without which may lead to a very hard situation.

wengzhenjie:
> I assume you may need a `pre-processor` to construct your SQL config without which may lead to a very hard situation.

I can't understand this.  I use GitHub API to get id.If I use `pre-processor`, I need to config the token. Can you tell me your logic about `pre-processor` ?

frank-zsy:
I mean how can you pass the repo id list into SQL renderer? You need to construct the render param string even you got all the repo id and org id list in json config.

wengzhenjie:
> I mean how can you pass the repo id list into SQL renderer? You need to construct the render param string even you got all the repo id and org id list in json config.

I pass repo_id in `having sentence` .
image

frank-zsy:
Need to figure out if this works in current render lib, the template engine may not be so powerful.

wengzhenjie:
> Need to figure out if this works in current render lib, the template engine may not be so powerful.

Sorry. I am not sure it works. I will fix it.

sqls/activity-repo-top-Chinese/pre-processor.js
```diff
@@ -0,0 +1,20 @@
...
+    let org_ids = new Array();
+    for (let i in config.orgs) {
+        org_ids.push(config.orgs[i].id);
+    }
+    let repo_ids = new Array();
+    for (let i in config.repos) {
+        repo_ids.push(config.repos[i].id);
+    }
+    let ids={
+        "actor_ids":actor_ids,
```
frank-zsy, line 15: You can not use array here, you need to construct a string.

wengzhenjie: OK.

frank-zsy: Actually I think it's not about the string or array, it's because wrong param in SQL template

sqls/activity-repo-top-Chinese/sql
```diff
@@ -0,0 +1,26 @@
...
+SUM(icc.count) AS issue_comment,SUM(oic.count) AS open_issue, SUM(opc.count) AS open_pull, SUM(rcc.count) AS pull_review_comment, SUM(mpc.count) AS merge_pull,
+SUM(contribute_list.commits) AS commits,SUM(contribute_list.additions) AS additions,SUM(contribute_list.deletions) AS deletions
+FROM
+    (SELECT
+    icc.repo_id AS repo_id,icc.repo_name AS repo_name, icc.actor_id AS actor_id, icc.count, oic.count, opc.count, rcc.count, mpc.count,
+    {{weight.issueCommentWeight}}*icc.count+{{weight.openIssueWeight}}*oic.count+{{weight.openPullWeight}}*opc.count+{{weight.pullReviewWeight}}*rcc.count+{{weight.mergePullWeight}}*mpc.count AS score,
+    mpc.commits AS commits,mpc.additions AS additions,mpc.deletions AS deletions,mpc.repo_language AS repo_language
+    FROM
+        (SELECT org_id,repo_id, actor_id,anyHeavy(repo_name) AS repo_name, COUNT(*) count FROM {{table}} WHERE type='IssueCommentEvent' AND action='created' GROUP BY org_id,repo_id, actor_id HAVING actor_id IN [{{actor_ids}}] OR repo_id IN [{{repo_ids}}] OR org_id IN [{{org_ids}}]) AS icc
```
frank-zsy, line 10: You should not use `repo_ids`, you should use `ids.repo_ids` instead.

frank-zsy (changes requested):

sqls/activity-repo-top-Chinese/sql
frank-zsy: Nope, should change to string first to see the result.

frank-zsy:
False approve.

/sql-run

PR #267, opened 2020-12-17
[Feature] Add pre processor

### LINKED ISSUE #265: [Refactor] We may need a pre-processor for SQL component
For now, we have `sql`, `manifest.json` and `post-processor.js` for every SQL component, the configs in `manifest.json` file is used to render SQL query and report text.

But as the template engine([pope](https://www.npmjs.com/package/pope)) we use is not powerful enough to handle complex data structure, so we may need a `pre-processor.js` file to process the configs to strings that can be directly used by SQL renderer which will give us more ability to form complex SQL.

Related to #251

frank-zsy:

Add pre processor to SQL component which will accept manifest config as param and can return an object for render SQL.

close #265

### DIFF
--- a/.github/hypertrons-components/auto_update_report/config.ts
+++ b/.github/hypertrons-components/auto_update_report/config.ts
@@ -51,6 +51,12 @@ export default class Config {
   sqlManifestFile: string;
 
   @configProp({
+    description: 'SQL pre processor file path',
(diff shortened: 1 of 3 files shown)


frank-zsy:
I will merge this PR for component test.

PR #271, opened 2020-12-18
Update REPORT_TEMPLATE.md

16647615268:
Adjusted the image format


frank-zsy:
Do we need HTML style image block? I think Markdown image in new line is enough for the report. WDYT?

16647615268:
The images here are not the same size and need to be in this format

frank-zsy:
OK, understood

PR #282, opened 2020-12-20
add Chinese company sql component

wengzhenjie:
#206 

add company info to config


wengzhenjie:
I guess the error may be related to the type of id on `join`.

frank-zsy:
Maybe, I think I should modify the server side to return the error detail instead of status 500.

frank-zsy:
I still think we can use a single component to generate Chinese and Chinese company status. We can return all repo data as result and group by company in post-processor, because I think the result may not be too large since there are only several fields for one repo.

wengzhenjie:
I see. I will change it.

wengzhenjie:
`post-processor` is required to return string. But I want to return list. How can I do for report_template?

frank-zsy:
Actually you can also return an object and use key-value in template file. like `return {"c": "a"}` and in template file you can use `slq.text.c` to get the value.

wengzhenjie:
OK. Thanks.

frank-zsy:
As we have no `pre-processor` implementation for CI part, I will merge this PR to test first.

/approve

Issue #276, opened 2020-12-19
[Feature] Developer analysis for case study.


frank-zsy:
We need a developer case study SQL component.

We can return the most active developer accounts for case study repos.

chenxi112:
Step1: According to the total activity of developers, we select the top 20 developers.
Step2: Based on the top 20 developers, We analyze the activity of each developer on different projects. As a developer may be involved in too many projects, I suggest that the top 10 projects be displayed in descending order according to their own activity on different projects.

frank-zsy:
I think the result is OK but we don't really need to split into 2 steps, the result could be gathered in one SQL I think.

chenxi112:
OK, Of course they will be gathered in one SQL.

PR #290, opened 2020-12-23
sqls: add sql of developer (#276)

chenxi112:
"sqls: add sql of developer (#276)"


frank-zsy:
Can you describe what the output is? So we can review the SQL component due to the description.

chenxi112:
It will output top 20 developers according to their total activity. And for each developer,  it will output the top 10 most active repositories, their activity, issue-comment,  open_issue, open_pull, pull_review_comment, merge_pull, pull comments, pull_additions and pull_deletions.

frank-zsy:
But I can not quite understand your SQL here. I think there are some points need to be considered.

- This a case study component, so we may not need to constrain the repo count, all repos in the case study need to be returned.
- For every repo in the case study, we may return top N most active developers in the repo during this year.
- Your SQL contains some errors and I can not understand the logic here, we may discuss this later in the meeting.

frank-zsy:
I am not sure this is correct or not, I can't tell from my eyes. I will merge this PR for testing as we don't have case study CI right now.

/approve

Issue #277, opened 2020-12-19
[Feature] Release analysis for case study


frank-zsy:
We need a release analysis SQL component for case study.

TieWay59:
First we can list out the latest release records with version code, publish time, and etc. (list about 10 items or more)
Base on the listed releases item, we can analyze the frequency of the releases, which can laterally tell more detail of project/community liveliness.

frank-zsy:
I think we don't need to list release records, we can use some statistics data instead. The latest release time and version should be included. And there are some other indicators like:

- How many times the project releases in 2020.
- Average release interval.
- Average release body string length.

You may add other indicators as you like too.

TieWay59:
OK, I'll add them in the sql components. 
Should I join them in one component or split indicators respectively?

frank-zsy:
All release indicators should be in one SQL component I think.

PR #296, opened 2020-12-25
sqls: add sql of release analysis (#277)

TieWay59:
This sql returns the total releases count, last release time, average interval and average body length of every repo.

To display the result directly, I build a simple data case as below:

图片

The result which this sql returns will be like below.

图片

I’m not so sure is this what we want in this feature or does this implementation fit the runtime environment. @frank-zsy

### DIFF
--- a/.autogit/autogit.json
+++ b/.autogit/autogit.json
@@ -1,0 +1,5 @@
+{
+  "updateInterval": 1800,
+  "logging": true,
+  "silent": true
+}
--- a/sqls/release-analysis/sql
+++ b/sqls/release-analysis/sql
@@ -1,0 +1,10 @@
+SELECT
+    repo_id,
+    COUNT(*) AS total_releases_count,
+    MAX(release_published_at) AS last_release_time,
+    365 / COUNT(*) AS average_interval,
+    AVG(LENGTH(release_body)) AS average_body_length
+FROM
+    {{table}}
+GROUP BY
+    repo_id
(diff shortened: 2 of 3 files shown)


.autogit/autogit.json
```diff
@@ -0,0 +1,5 @@
+{
+  "updateInterval": 1800,
```
frank-zsy, line 2: You may need to remove this file from the commit.

sqls/release-analysis/sql
```diff
@@ -0,0 +1,10 @@
+SELECT
+    repo_id,
+    COUNT(*) AS total_releases_count,
+    MAX(release_published_at) AS last_release_time,
+    365 / COUNT(*) AS average_interval,
+    AVG(LENGTH(release_body)) AS average_body_length
+FROM
+    {{table}}
```
frank-zsy, line 8: The release analysis should be a case study SQL component which should be in case study folder and the result should be for every single repo but not in a total aggregation.

frank-zsy: My bad, actually you `GROUP BY repo_id` will split into different repos, but without `WHERE` clause you can not get correct result of release status. `COUNT(*)` will count all log counts of a repo in a year.

TieWay59: I got what you mean, I think I still have no enough insight on the database structure. So I directly mistake all records for release item. Should I close this PR or keep on fix this? since this feature are included on another component.

frank-zsy: Can I close this PR and add the function myself?

TieWay59: Yes and sorry for all the inconvenience and delay I have caused.

frank-zsy: No problem.

PR #329, opened 2021-01-05
[Fix] Several changes to Chinese repo 

frank-zsy:

Fix several things about Chinese repo SQL component.

For hypertrons `auto_update_report` component
- Need to pass self component config into post processor as the third param, the second param is the global default render params.

For Chinese SQL component:
- As returned `id` columns are in string type, need to change `org_id_company_map` and `repo_id_company_map` key to string type, otherwise will not find the correct company.
- As returned integer number columns are in string type, need to use `parseInt` to change into `int` because need to calculate later.
- the `company_info_list` is not in correct format, need to change into an array with array elements and the `activity` needs to be fixed to 2 digits as add operation will do some tricks to float value.

### DIFF
--- a/.github/hypertrons-components/auto_update_report/index.lua
+++ b/.github/hypertrons-components/auto_update_report/index.lua
@@ -49,7 +49,7 @@ sched(compConfig.schedName, compConfig.sched, function ()
             ['query'] = sql
           }
         })
-        local renderText = runJsCode(postProcessor, string2table(requestRes).data, compConfig.defaultRenderParams)
+        local renderText = runJsCode(postProcessor, string2table(requestRes).data, compConfig.defaultRenderParams, manifest.config)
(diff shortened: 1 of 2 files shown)


frank-zsy:
@wengzhenjie Can you review this one?

frank-zsy:
> /sql-run

You don't need to run this SQL because CI still not support hypertrons and pre processor, only look into the changes and see if they make sense.

wengzhenjie (approved):

wengzhenjie (approved):

wengzhenjie:
These changes are useful. Company_info_list actually is two-dimensional array. And some fields also need to be converted. Thank you fixing them.

frank-zsy:
Alright, I will force merge this PR to see the result.

PR #331, opened 2021-01-06
Ben s move patch 1

BenS-Move:


BenS-Move:
Through the analysis of the robot account, I summarized the relevant attribute information of the robot account.

BenS-Move:
Through the analysis of the robot account, I summarized the relevant attribute information of the robot account.

frank-zsy:
Thanks for the PR.

- Why do you update two files with same content in this PR?
- I think the robot analysis content should be in the REPORT_TEMPALTE.md file right under developers analysis part, but not in a single file, WDYT?

PR #336, opened 2021-01-07
[Fix] Working hour for case study

frank-zsy:

Fix some errors about the working hour case study component.

- Use `k` instead of `i` as loop param because `i` is used in inner loop which will cause a infinite loop here.
- Add repo name to render text to indicate which repo the image belongs to.
- Use string interpolation instead of string join with `+`.


frank-zsy:
@gymgym1212 Some changes to your component.

PR #337, opened 2021-01-07
[Fix] top developer analysis for case study

frank-zsy:

Fix few things about top developer case study component.

- Name change, the `topN` should be in config so `20` should not be included in the component name.
- Add self component config as third param to `post-processor`.
- Fix SQL and `post-processor`, the SQL should not like global analysis, it should be just group by repo and actor, and we can use a `minActivity` to filter some developers first.


frank-zsy:
I will merge this PR to test.

PR #345, opened 2021-01-07
[Feature] Add release and issue resolve period to activity global

frank-zsy:

As release status and issue resolve period can be several lines about a repo, so we can just add the info into repo activity SQL component.


frank-zsy:
Global release and issue resolve period will exceed the max memory limit of Clickhouse server, close this and add to case study.

PR #348, opened 2021-01-07
add company repo_count&developer_count

wengzhenjie:


frank-zsy:
@sunshinemingo As the CI component does not support pre processor yet, the /sql-run won't return correct result, I will add it first.

sunshinemingo:
> @sunshinemingo As the CI component does not support pre processor yet, the /sql-run won't return correct result, I will add it first.

OK, I get it.

frank-zsy:
The returned SQL result is too long to comment to a PR, I will merge the PR, I think it is OK.

/approve

PR #362, opened 2021-01-10
[bug]Modify the error serial number

QuakeWang:
I found on the contributing guide page that some serial numbers are incorrect on the Chinese page and the English page, so I submitted this pr.


frank-zsy:
Thanks for your fix
/approve

PR #387, opened 2021-01-20
update Chinese_repo config

wengzhenjie:
I correct some errors about repos. I think we should check this config. Maybe there are some actors and repos not in this.


frank-zsy:
Thanks for the fix.

/approve

PR #525, opened 2021-04-15
[Docs] Modify data description

frank-zsy:

While we modify our architecture of the project, we also modify the data structure of the database. Includes:

- Add `pull_request_commit_sha` column which can be used to relate `PullRequestEvent` and `PushEvent`.
- Change array items from JSON string into Clickhouse `Nested` type for analysis
  - `issue_labels` which contains `name`, `color`, `default`, `description`
  - `issue_assignees` which contains `id`, `login`
  - `push_commits` which contains `name`, `email`, `message`
  - `gollum_pages` which contains `page_name`, `title`, `action`
  - `release_assets` which contains `name`, `uploader_id`, `uploader_login`, `content_type`, `state`, `size`, `download_count`

### DIFF
--- a/docs/csv/data_description.csv
+++ b/docs/csv/data_description.csv
@@ -1,128 +1,143 @@
 name,type,description,payload_type
-id,UInt64,uique identity of this event instance,general
-type,String,"event instance type,event instances of different event type have different payload from GitHub API",general
-action,String,action in the payload of event instance,general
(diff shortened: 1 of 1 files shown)


frank-zsy:
With `Nested` type we can analyze array data easier than parse JSON string like: 

image

Issue #677, opened 2021-11-23
[Doc] We need an issue template file to help others to submit labeled data


frank-zsy:
We may need an issue template to help developers to submit their data like company repos, repos that they donate to foundations and community with multiple repos. Something like these.

They may submit the data as markdown format list and we can add them into config file ourselves.

frank-zsy:
Can you do this one? @gymgym1212

gymgym1212:
> Can you do this one? @gymgym1212

Yes

gymgym1212:
I think the table syntax of markdown is too cumbersome, which will bring great obstacles to users to fill in. How about changing the syntax to something like a yaml file?
```
orgs:
  - 1234567 # org_name
    - ...
repos:
  - 7654321 # repo_name
  - ...
``` 

In addition, do users know how to get the ID of the organization or repo? Should we also add a paragraph of guidance to tell users how to obtain IDs?

frank-zsy:
Yes, markdown table is too complex for others, bullet points is good enough but we don't really need to left the two whitespace to fit the yaml syntax, I think plain markdown list is good enough for this.

PR #678, opened 2021-11-24
Add new issue templates for submitting labeled data and ordinary issue

gymgym1212:
There are two new template files users can choose.

#677

### DIFF
--- a/.github/ISSUE_TEMPLATE/ordinary_issue.md
+++ b/.github/ISSUE_TEMPLATE/ordinary_issue.md
@@ -1,5 +1,5 @@
 ---
-name: Ordinary Issue
+name: Ordinary issue
 about: Everything you want to tell us
 ---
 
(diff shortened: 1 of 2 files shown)


gymgym1212:
https://github.com/gymgym1212/open-digger/issues/new?assignees=&labels=&template=submit_labeled_data.md

frank-zsy:
If we add an issue template, we may need to consider ourselves as normal developers and stand in their position.

So the content of the issue should be from a developer which may contains like `"I want to add some labeled data into OpenDigger to help us for our community analysis"`.

And the information we want developers aware should be annotated like you see here: https://github.com/microsoft/vscode/blob/main/.github/ISSUE_TEMPLATE/bug_report.md?plain=1

gymgym1212:
Yes, this is my problem, and I will fix it as soon as possible.

frank-zsy:
Thx, I think this template is enough for most scenarios.

/approve

PR #761, opened 2022-03-05
Add New Region Label Data America.yml

xiaoya-Esther:


labeled_data/companies/vmware/index.yml
```diff
@@ -1,9 +1,15 @@
 name: VMWare
 type: Company
 data:
+  github_repo:
```
frank-zsy, line 4: This empty field should be removed too.

xiaoya-Esther: OK

PR #785, opened 2022-03-20
[Docs] Update qrcode

frank-zsy:

Update the QRCode since we already have more than 100 members in the group, users can not just scan the QRCode and join the group by themselves, so I will let them add me as friend and I will invite them into the group manually.

Another advantage is: the QRCode of personal account will not expire like a chat group one.


will-ww:
> Signed-off-by: Frankzhaopku [syzhao1988@126.com](mailto:syzhao1988@126.com)
> 
> Update the QRCode since we already have more than 100 members in the group, users can not just scan the QRCode and join the group by themselves, so I will let them add me as friend and I will invite them into the group manually.
> 
> Another advantage is: the QRCode of personal account will not expire like a chat group one.

Great, many thanks~

PR #800, opened 2022-03-30
Add more GitHub organization accounts of QingCloud

LinuxSuRen:


frank-zsy:
Thanks to perfect the data.

/approve

frank-zsy:
@LinuxSuRen The online data has been updated. QingCloud activity changes from No.8 to No.6 and influence changes from No.10 to No.8 in 2021 among Chinese companies. Check out here: https://open-leaderboard.x-lab.info/

LinuxSuRen:
Thanks for your quick response.

I guess some other companies have multiple GitHub organizations as well. I'm thinking if it's possible to have a detail page for the company. We can show the organizations of a particular company. It might be helpful.

frank-zsy:
> Thanks for your quick response.
> 
> I guess some other companies have multiple GitHub organizations as well. I'm thinking if it's possible to have a detail page for the company. We can show the organizations of a particular company. It might be helpful.

Yes, great advise, right now all the data is yml files and maybe we should add a webpage to show all the labeled data by table so users can find out the missing data easily.

I will rise an issue about this and add a script to generate a markdown page due to the current label data.

frank-zsy:
But there is a tricky thing that as we only use repo or organization ids to identify, the names of the item (which can changed from time to time) are in the comments which can not be extracted from the raw data. Is there any suggestion here? Or maybe we should change the yaml data format to also save the names in label data? @LinuxSuRen

LinuxSuRen:
I'm not sure if I understand it correctly. Do you mean that the organization name of a company might be changed in some time?

If I understand it correctly, the REAL id of an organization cannot change. For example, the id of `kubesphere` is `37326490`.

I don't know if there is a way to find the URL of an oranization by its id.

frank-zsy:
Yes, so we need to generate the doc in a GitHub action or by Hypertrons to fetch the name automatically while generating the doc. I will do that~Thanks

PR #840, opened 2022-05-26
Add a new GitHub org which belong to QingCloud

LinuxSuRen:
See also the following screenshots:

image
image

### DIFF
--- a/labeled_data/companies/qingcloud/index.yml
+++ b/labeled_data/companies/qingcloud/index.yml
@@ -9,4 +9,4 @@ data:
     - 86996287 # OpenELB
     - 31385114 # radondb
     - 28995775 # QingCloudAppcenter
-    - 90166053f # quanxiang-cloud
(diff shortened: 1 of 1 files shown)


labeled_data/companies/qingcloud/index.yml
```diff
@@ -9,3 +9,4 @@ data:
     - 86996287 # OpenELB
     - 31385114 # radondb
     - 28995775 # QingCloudAppcenter
+    - 90166053f # quanxiang-cloud
```
frank-zsy, line 12: I think you added an extra `f` in the end of the id

frank-zsy (changes requested):

labeled_data/companies/qingcloud/index.yml
LinuxSuRen: Good catch!

frank-zsy:
Thanks for the data enhancement

/approve

PR #892, opened 2022-06-26
metric: add code change commits metric

xiaoya-yaya:
So far I only implemented the total commits count.

Refer to the description of this metric, it can be filtered by actors: `committers or normal contributors`, or tags: `bugfix or feature`, of which the data is hard to obtain right now.

I think we can also implement the filter of commits by `direct push`(total_commits - accepted PRs) or `change requests.`


xiaoya-yaya (commented):
Eveything looks good to me except a small detail: if the demo displays the curve from 2015-2021, the title of the chart should be changed to `by year` instead of `by month`

frank-zsy:
Yes, I have fixed the demo description.

Issue #584, opened 2021-06-02
[Metrics] Change Request Acceptance Ratio


xiaoya-yaya:
**Description**
Each change request can be in one of three states: **open, merged (accepted), and closed without merge
(declined)**. This metric measures the ratio of change requests merged (accepted) vs change requests
closed without being merged.

**Instance of Implementation**
![image](https://user-images.githubusercontent.com/56542972/120425874-97458a00-c3a1-11eb-9b87-1ea990942ba2.png)

**Resource**
CHAOSS: Evolution -> Code Development Activity

longyanz:
Hi, i'm trying to implement this mertic. I calculate the ratio in two ways，the first is ratio =  pr closed without being merged/pr merged, the second is ratio = pr merged/pr closed(including pr closed without being merged and pr merged)。The result is shown below 
![image](https://user-images.githubusercontent.com/91717658/199640629-9a863300-27b9-43ae-926f-1353b0512bfa.png)
![image](https://user-images.githubusercontent.com/91717658/199640643-99515068-5b46-424c-8d29-16a2e42b03bf.png)

But i think the result not represented  the ratio of change requests merged (accepted) vs change requests closed without being merged well. Maybe I should use a histogram，just like the Instance of Implementation？Can anyone give me some advice？
@xgdyp

xiaoya-yaya:
I agree that the histogram is a better presentation here. 

As for the value itself, I think the second way (merged/all closed) represents the meaning of this metric as "the ratio of accepted PRs among all the PRs"

frank-zsy:
closed by #1097

PR #936, opened 2022-08-03
Metric: Change Requests Accepted Ratio

longyanz:
Hi, I am trying to implement the metric : change requests accepted ratio #584  . But I got the following error when running the local test. Is the problem with the "countIF" codes ?
51c6b272aa7b7defaac29339a5cf0f2


src/metrics/chaoss.ts
```diff
@@ -295,3 +295,44 @@ FORMAT JSONCompact`;
...
+  const whereClauses: string[] = ["type = 'PullRequestEvent' AND action = 'closed'"];
+  const repoWhereClause = getRepoWhereClauseForClickhouse(config);
+  if (repoWhereClause) whereClauses.push(repoWhereClause);
+  whereClauses.push(getTimeRangeWhereClauseForClickhouse(config));
+
+  const sql = `
+SELECT
+  id,
+  argMax(name, time) AS name,
+  ${getGroupArrayInsertAtClauseForClickhouse(config, { key: 'accepted'})}
```
frank-zsy, line 310: You should add a comma at the tail of the line.

longyanz: Thanks！

PR #937, opened 2022-08-03
metric: issue duration resolution

### LINKED ISSUE #590: [Metrics] Issue Resolution Duration
**Description**
- This metric is an indication of how long an issue remains open, on average, before it is closed.
- For issues that were reopened and closed again, only **the last close date** is relevant for this metric.

**Implementation**
- Issue Resolution Duration = Timestamp of issue closed - Timestamp of issue opened
- Average. Average amount of time (in days, by default) for an issue in the repository to be closed.

**Resource**
CHAOSS: Evolution -> Issue Resolution
https://chaoss.community/metric-issue-resolution-duration/

frank-zsy:

close #590 

Add issue resolution duration metric from CHAOSS.


frank-zsy:
@longyanz Time range related metric could be quite tricky, you can review on this one.

frank-zsy:
I will merge this PR first, I have tested it and add demo to notebook, but I think you can check the implementation and learn how to handle time range related metrics. @longyanz @xiaoya-yaya

longyanz:
Thanks！

src/metrics/chaoss.ts
```diff
@@ -295,3 +295,66 @@ FORMAT JSONCompact`;
...
+  by: 'open' | 'close';
+  type: 'avg' | 'median';
+  unit: 'week' | 'day' | 'hour' | 'minute';
+}
+export const chaossIssueResolutionDuration = async (config: QueryConfig<IssueResolutionDurationOptions>) => {
+  config = getMergedConfig(config);
+  const whereClauses: string[] = ["type = 'IssuesEvent'"];
+  const repoWhereClause = getRepoWhereClauseForClickhouse(config);
+  if (repoWhereClause) whereClauses.push(repoWhereClause);
+
```
frank-zsy, line 309: Do not add time range condition here, since we need to search the whole database for issue open and close time. If users config `by` to `open`, then `opened_at` should be in the config time range, but not necessary for `closed_at`, so the condition should be after `WHERE` clause and in `HAVING` clause to filter the result.

src/metrics/chaoss.ts
```diff
@@ -295,3 +295,66 @@ FORMAT JSONCompact`;
...
+      org_id,
+      argMax(org_login, created_at) AS org_login,
+      issue_number,
+      argMaxIf(action, created_at, action IN ('opened', 'closed' , 'reopened')) AS last_action,
+      maxIf(created_at, action = 'opened') AS opened_at,
+      maxIf(created_at, action = 'closed') AS closed_at
+    FROM github_log.events
+    WHERE ${whereClauses.join(' AND ')}
+    GROUP BY repo_id, org_id, issue_number
+    HAVING ${byCol} >= toDate('${config.startYear}-${config.startMonth}-1') AND ${byCol} < toDate('${endDate.getFullYear()}-${endDate.getMonth() + 1}-1') AND last_action='closed'
```
frank-zsy, line 342: - Filter time range here to avoid losing data for `opened_at` or `closed_at` timestamp in the original query.
- Check `last_action='closed'` to make sure you find a closed issue rather than a reopened issue.

src/metrics/chaoss.ts
```diff
@@ -295,3 +295,66 @@ FORMAT JSONCompact`;
...
+      argMax(repo_name, created_at) AS repo_name,
+      org_id,
+      argMax(org_login, created_at) AS org_login,
+      issue_number,
+      argMaxIf(action, created_at, action IN ('opened', 'closed' , 'reopened')) AS last_action,
+      maxIf(created_at, action = 'opened') AS opened_at,
+      maxIf(created_at, action = 'closed') AS closed_at
+    FROM github_log.events
+    WHERE ${whereClauses.join(' AND ')}
+    GROUP BY repo_id, org_id, issue_number
```
frank-zsy, line 341: `org_id` column will be used in `WHERE` condition clause, so need it in `GROUP BY` clause and do not add `any` or `argMax` to group it.

src/metrics/chaoss.ts
xiaoya-yaya: I don't quite get the meaning of the filter `by open_at` or `by close_at`, shouldn't the **issue resolution duration** be the avg/median duration timeframe from when issues are **opened** to when issues are **closed**?

What does it mean if I use the filter `close`?

frank-zsy: We need to aggregate issues into different timeframe. Like if you want to get the issue resolution duration in 2021, so you need to point out whether you mean all the issues opened in 2021 or closed in 2021.

The parameter is also mentioned in the original doc: https://chaoss.community/metric-issue-resolution-duration/

Issue #588, opened 2021-06-02
[Metrics] Issue Age


xiaoya-yaya:
**Description**
- How long have open issues been left open?
- If an issue has been closed but re-opened again within that period will be considered as having remained open since its **initial opening date.**

**Objectives**
- When the issue age is increasing, identify the oldest open issues in a project to gain insight as to why they have been open for an extended period of time. 
- Additionally, to understand how well maintainers are resolving issues and how quickly issues are resolved.

**Implementation**
- Average. Average age of all open issues.
- Median. Median age of all open issues.
- Filter by: module, tags/labels on issue
![image](https://user-images.githubusercontent.com/56542972/120429115-71bb7f00-c3a7-11eb-935e-166512604651.png)
![image](https://user-images.githubusercontent.com/56542972/120429284-b47d5700-c3a7-11eb-9d4f-48979f529a70.png)

**Resource**
CHAOSS: Evolution -> Issue Resolution
https://chaoss.community/metric-issue-age/

frank-zsy:
In the implementation of this metric, I found it is very tricky, especially about how to handle time range related functions.

As the description of the metric -- `How long have open issues been left open?` -- So actually the metric should not have a start time and an end time as parameter but just a time point, and returns how long have the open issues been left open, and the issues should be in open state at that time point.

For example, if we are going to get metric for a repo in 2022.1 - 2022.12 by quarter, so we will get 4 data points, each one contains the data about how long have the issues been left open until the end of the quarter, and the open time should not be limited. This really makes the metric hard to implement.

birdflyi:
A query of `Issue Age` for one issue might have those states: 
 1. query time (end time: default now), 
 2. the first open time (initial opening time), 
 3. the last closed time (when issue is closed at the point of query time),
 4. active time intervals (for some issue may be closed or reopened), 
 5. inactive time intervals (the accumulate time intervals of closed-reopened), 
 6. how many times of inactive-active breaks,  # this may be a redundant property
 7. active state (open, reopened or closed)
 
when I input an `issue_number` and an `end_time{default now}` and a parameter `include_closed2reopened{default True}`, I'd like to get those results as the issue age: 
select all the records about `issue number` created before `query point in time` from tables in our databases, return
- include_closed2reopened=True
  - None  # for issues not exist or cannot be retrieved from our databases.
  - Non-negative Number of 3. the last closed time - 2. the first open time  # for issues closed.
[...]

PR #949, opened 2022-08-17
Feat: Implementing Issue Age Metric

xiaoya-yaya:
I'm implementing metric #588 and running into some issues.

The error message is :

`Error: Missing columns: 'created_at' while processing query: 'SELECT toStartOfMonth(created_at) AS time, arrayJoin(multiIf((org_id IN (476009, 2976699, 3536625, 4542585, 4175954, 2718604, 4028872, 32504, 4764587, 5420974, 1440554)) OR (repo_id IN (6898381, 6898387, 5611582)), ['Adobe']......GROUP BY repo_id, org_id, issue_number HAVING (opened_at < toDate('2016-1-1')) AND ((last_action = 'opened') OR (last_action = 'reopened'))) GROUP BY id, time ORDER BY issue_age DESC LIMIT 10 BY time', required columns: 'org_id' 'repo_id' 'opened_at' 'created_at', source columns: 'last_action' 'org_login' 'issue_number' 'opened_at' 'org_id' 'repo_name' 'repo_id'`

### DIFF
--- a/src/metrics/chaoss.ts
+++ b/src/metrics/chaoss.ts
@@ -358,3 +358,62 @@ FORMAT JSONCompact`;
     }
   });
 };
+
+interface IssueAgeOptions {
+  type: 'avg' | 'median';
+  unit: 'week' | 'day' | 'hour' | 'minute';
+}
+export const chaossIssueAge = async (config: QueryConfig<IssueAgeOptions>) => {
+  config = getMergedConfig(config);
+  const whereClauses: string[] = ["type = 'IssuesEvent'"];
+  const repoWhereClause = getRepoWhereClauseForClickhouse(config);
+  if (repoWhereClause) whereClauses.push(repoWhereClause);
+
+  const endDate = new Date(`${config.endYear}-${config.endMonth}-1`);
+  endDate.setMonth(config.endMonth);
+  
+  let type = filterEnumType(config.options?.type,['avg','median'],'avg');
+  let unit = filterEnumType(config.options?.unit,['week','day','hour','minute'],'day');
+
+  const sql = `
+SELECT 
+  id, 
+  argMax(name, time) AS name,
+  ${getGroupArrayInsertAtClauseForClickhouse(config, {key:`issue_age`, defaultValue: 'NaN'})}
+FROM
+(
+  SELECT
+    ${getGroupTimeAndIdClauseForClickhouse(config, 'repo')},
+    round(${type}(dateDiff('${unit}', opened_at, '${endDate.getFullYear()}-${endDate.getMonth()+1}-1')),${config.percision}) AS issue_age
+  FROM
+  (
+    SELECT
+      repo_id,
+      argMax(repo_name, created_at) AS repo_name,
+      org_id,
+      argMax(org_login, created_at) AS org_login,
+      issue_number,
+      argMaxIf(action, created_at, action IN ('opened', 'reopened', 'closed')) AS last_action,
+      maxIf(created_at, action = 'opened') AS opened_at
+    FROM github_log.events
+    WHERE ${whereClauses.join('AND')}
+    GROUP BY repo_id, org_id, issue_number
+    HAVING opened_at < toDate('${endDate.getFullYear()}-${endDate.getMonth()+1}-1') AND (last_action='opened' OR last_action='reopened')
+  )
+  GROUP BY id,time
+  ${config.limit > 0 ? `ORDER BY issue_age ${config.order} LIMIT ${config.limit} BY time` : ''} 
+)
+GROUP BY id
+ORDER BY issue_age[-1] ${config.order}
+FORMAT JSONCompact`;
+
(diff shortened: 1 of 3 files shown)


xiaoya-yaya:
The logic is finding all the opening issues before the selected end date, and calculating the time duration from the issue opened date to the selected end date.

I'll add some comments to the code.

src/metrics/chaoss.ts
```diff
@@ -358,3 +358,62 @@ FORMAT JSONCompact`;
...
+  const sql = `
+SELECT 
+  id, 
+  argMax(name, time) AS name,
+  ${getGroupArrayInsertAtClauseForClickhouse(config, {key:`issue_age`, defaultValue: 'NaN'})}
+FROM
+(
+  SELECT
+    ${getGroupTimeAndIdClauseForClickhouse(config, 'repo')},
+    round(${type}(dateDiff('${unit}', opened_at, '${endDate.getFullYear()}-${endDate.getMonth()+1}-1')),${config.percision}) AS issue_age
```
xiaoya-yaya, line 387: Calculate the time period from issues opened to the selected end time.

src/metrics/chaoss.ts
```diff
@@ -358,3 +358,62 @@ FORMAT JSONCompact`;
...
+      argMax(repo_name, created_at) AS repo_name,
+      org_id,
+      argMax(org_login, created_at) AS org_login,
+      issue_number,
+      argMaxIf(action, created_at, action IN ('opened', 'reopened', 'closed')) AS last_action,
+      maxIf(created_at, action = 'opened') AS opened_at
+    FROM github_log.events
+    WHERE ${whereClauses.join('AND')}
+    GROUP BY repo_id, org_id, issue_number
+    HAVING opened_at < toDate('${endDate.getFullYear()}-${endDate.getMonth()+1}-1') AND (last_action='opened' OR last_action='reopened')
```
xiaoya-yaya, line 401: Find all the issues that are remaining open(including reopened issues) before the selected end date.

src/metrics/chaoss.ts
```diff
@@ -358,3 +358,62 @@ FORMAT JSONCompact`;
...
+
+  const sql = `
+SELECT 
+  id, 
+  argMax(name, time) AS name,
+  ${getGroupArrayInsertAtClauseForClickhouse(config, {key:`issue_age`, defaultValue: 'NaN'})}
+FROM
+(
+  SELECT
+    ${getGroupTimeAndIdClauseForClickhouse(config, 'repo')},
```
frank-zsy, line 386: The `getGroupTimeAndIdClauseForClickhouse ` functions has another parameter which is the name of the time column. It will use `created_at` by default if not set which covers most situation.

So you should call `getGroupTimeAndIdClauseForClickhouse(config, 'repo', 'opened_at')` since the inner select table has no `created_at` column and issues should group by open time.

src/metrics/chaoss.ts
```diff
@@ -358,3 +358,62 @@ FORMAT JSONCompact`;
...
+  const sql = `
+SELECT 
+  id, 
+  argMax(name, time) AS name,
+  ${getGroupArrayInsertAtClauseForClickhouse(config, {key:`issue_age`, defaultValue: 'NaN'})}
+FROM
+(
+  SELECT
+    ${getGroupTimeAndIdClauseForClickhouse(config, 'repo','opened_at')},
+    round(${type}(dateDiff('${unit}', opened_at, toDate('${endDate.getFullYear()}-${endDate.getMonth()+1}'))),${config.percision}) AS issue_age
```
xiaoya-yaya, line 387: Now the error message is :
`Error: Syntax error: failed at position 7046 ('Jan') (line 9, col 47): Jan 01 2016 00:00:00 GMT+0000 (GMT))),2) AS issue_age  FROM  (    SELECT      repo_id,      argMax(repo_name, created_at) AS repo_name,      org_id,     .....`

I think it's due to the `startdate` and `enddate` parameters for the `dateDiff` function are in different datetime forms. For example, opened_at is something like `2015-01-29 20:11:07`, while the end date is something like `2015-01`

xiaoya-yaya: Do we have a better solution for this?

frank-zsy: The problem here is not format or data type difference. Actually the `startdate` and `enddate` can be `Date` or `DateTime`.

The `opened_at` is selected from the `created_at` column so it is `DateTime` type.

You use `toDate` to construct the `enddate` so it will be `Date` type. But it is OK to get `dateDiff` result for a `Date` param and a `DateTime` param.

So the real problem is you should change `toDate('${endDate.getFullYear()}-${endDate.getMonth()+1}')` to `toDate('${endDate.getFullYear()}-${endDate.getMonth()+1}-1')` which will change `2015-1` to `2015-1-1`, so the `toDate` function can understand what you want.

xiaoya-yaya:
I will close this PR first.
image

PR #952, opened 2022-08-24
Feat: Add Python kernel

Best-Eminem:
I have done the python kernel, but only test some cases using `Clickhouse` mirror. `Neo4j` API haven't been tested ever.

### DIFF
--- a/src/db/clickhouse.py
+++ b/src/db/clickhouse.py
@@ -1,0 +1,29 @@
+import sys
+import os
+from easydict import EasyDict
+from matplotlib.pyplot import get
+sys.path.append(os.getcwd())
+from src.config import getConfig
+from clickhouse_driver import Client
+
+_client = None
+def getClient(): 
+    global _client
+    if _client == None: 
+        # _client = new ClickHouse((await getConfig()).db.clickhouse);
+        clickhouse_config = EasyDict(EasyDict(getConfig()).db.clickhouse)
+        # print(clickhouse_config)
+        host = clickhouse_config.host
+        port = clickhouse_config.port
+        user = clickhouse_config.user
+        password = clickhouse_config.password
+        _client = Client(host,port)
+    return _client
+
+def query(q):
+    client = getClient()
+    return client.execute(q)
+def query_dataframe(q):
+    client = getClient()
+    return client.query_dataframe(q)
+
--- a/src/db/clickhouse.py
+++ b/src/db/clickhouse.py
@@ -2,17 +2,14 @@ import sys
 import os
 from easydict import EasyDict
 from matplotlib.pyplot import get
-sys.path.append(os.getcwd())
-from src.config import getConfig
+from config import getConfig
 from clickhouse_driver import Client
 
 _client = None
 def getClient(): 
     global _client
     if _client == None: 
-        # _client = new ClickHouse((await getConfig()).db.clickhouse);
         clickhouse_config = EasyDict(EasyDict(getConfig()).db.clickhouse)
-        # print(clickhouse_config)
         host = clickhouse_config.host
         port = clickhouse_config.port
         user = clickhouse_config.user
@@ -23,7 +20,7 @@ def getClient():
 def query(q):
     client = getClient()
     return client.execute(q)
-def query_dataframe(q):
+def queryDataframe(q):
     client = getClient()
     return client.query_dataframe(q)
 
--- a/src/db/clickhouse.py
+++ b/src/db/clickhouse.py
@@ -9,12 +9,8 @@ _client = None
 def getClient(): 
     global _client
     if _client == None: 
-        clickhouse_config = EasyDict(EasyDict(getConfig()).db.clickhouse)
-        host = clickhouse_config.host
-        port = clickhouse_config.port
-        user = clickhouse_config.user
-        password = clickhouse_config.password
-        _client = Client(host,port)
+        config = EasyDict((getConfig()).get('db').get('clickhouse'))
+        _client = Client(config.host, config.port, config.database, config.user, config.password)
     return _client
 
 def query(q):
--- a/src/db/neo4j_driver.py
+++ b/src/db/neo4j_driver.py
@@ -15,4 +15,4 @@ def getClient():
       
 def query(query_sql):
     result = getClient().run(query_sql) # return a cursor object
-    return result
+    return result.data()  # transform cursor to list
--- a/src/db/neo4j_driver.py
+++ b/src/db/neo4j_driver.py
@@ -1,0 +1,27 @@
+from py2neo import Node, Relationship, Graph
+# import parser = require('parse-neo4j');
+import sys
+import os
+from easydict import EasyDict
+sys.path.append(os.getcwd())
+from src.config import getConfig
+
+_driver = None
+
+def getClient():
+    global _driver
+    if _driver == None:
+        neo4j_config = EasyDict(EasyDict(getConfig()).db.neo4j)
+        _driver = Graph(neo4j_config.host)
+    return _driver
+      
+def query(query_sql):
+    print(query_sql)
+    result = getClient().run(query_sql) #返回的结果是一个cursor对象
+    return result
+
+# export function parseRecord<T = any>(record: any): T {
+#   return parser.parseRecord(record)._fields[0];
+# }
+# resultaa = query("MATCH (r:Repo{name: 'pingcap/tidb'}) RETURN r")
+# print(resultaa)
--- a/src/db/neo4j_driver.py
+++ b/src/db/neo4j_driver.py
@@ -1,10 +1,8 @@
 from py2neo import Node, Relationship, Graph
-# import parser = require('parse-neo4j');
 import sys
 import os
 from easydict import EasyDict
-sys.path.append(os.getcwd())
-from src.config import getConfig
+from config import getConfig
 
 _driver = None
 
@@ -16,12 +14,5 @@ def getClient():
     return _driver
       
 def query(query_sql):
-    print(query_sql)
-    result = getClient().run(query_sql) #返回的结果是一个cursor对象
+    result = getClient().run(query_sql) # return a cursor object
     return result
-
-# export function parseRecord<T = any>(record: any): T {
-#   return parser.parseRecord(record)._fields[0];
-# }
-# resultaa = query("MATCH (r:Repo{name: 'pingcap/tidb'}) RETURN r")
-# print(resultaa)
--- a/src/metrics/activity_openrank.py
+++ b/src/metrics/activity_openrank.py
@@ -21,7 +21,7 @@ OPEN_PULL_WEIGHT = 3
 REVIEW_COMMENT_WEIGHT = 4
 PULL_MERGED_WEIGHT = 2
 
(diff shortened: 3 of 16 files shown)


src/metrics/activity_openrank.py
```diff
@@ -0,0 +1,239 @@
...
+from functools import cmp_to_key
+import numpy as np
+
+ISSUE_COMMENT_WEIGHT = 1
+OPEN_ISSUE_WEIGHT = 2
+OPEN_PULL_WEIGHT = 3
+REVIEW_COMMENT_WEIGHT = 4
+PULL_MERGED_WEIGHT = 2
+
+def get_RepoActivityOrOpenrank(config, type='activity'):
```
frank-zsy, line 24: We should keep an identical function style, underscore or camel. But not both at same time. Did you do this for a particular reason?

frank-zsy:
Do you test the codes in a notebook? Do these codes run in a common Python kernel notebook? Do we need to make a new notebook image with some libraries installed, like `plotly`?

Best-Eminem:
> 

I test the codes in python file but not yet in notebook, maybe I can create a notebook named 'handbook_py' for convenience?

src/metrics/activity_openrank.py
Best-Eminem: > We should keep an identical function style, underscore or camel. But not both at same time. Did you do this for a particular reason?

sorry, I will check all functions.

frank-zsy:
> 

Yes, actually I think when Python kernel is ready, we can remove all Node.js related files since it is not quite friendly to data science students and developers. So maybe `handbook_py` first and then replace the old files in the future.

Best-Eminem:
> > 
> 
> Yes, actually I think when Python kernel is ready, we can remove all Node.js related files since it is not quite friendly to data science students and developers. So maybe `handbook_py` first and then replace the old files in the future.

I have finished a notebook named `handbook_py`, and I use `Plotly` for Python to draw the charts.

frank-zsy:
Thanks, I will check it out. And do I need to install dependencies myself?

Best-Eminem:
> Thanks, I will check it out. And do I need to install dependencies myself?

I recommend installing these dependencies with `pip` based on `Anaconda` environment.

frank-zsy:
I think we need more documentation about the environment setup or a docker image maybe better like Node.js kernel. Since we don't sure the Python version for each developer and the the level they master Python. For instance, I am not quite familiar with Python.

What Python version do you use? Python 2.7 or Python 3?

Best-Eminem:
> I think we need more documentation about the environment setup or a docker image maybe better like Node.js kernel. Since we don't sure the Python version for each developer and the the level they master Python. For instance, I am not quite familiar with Python.
> 
> What Python version do you use? Python 2.7 or Python 3?

ok, I will make a docker image for python kernel.

Best-Eminem:
> I think we need more documentation about the environment setup or a docker image maybe better like Node.js kernel. Since we don't sure the Python version for each developer and the the level they master Python. For instance, I am not quite familiar with Python.
> 
> What Python version do you use? Python 2.7 or Python 3?

I have make a docker file for you to build. Please use `docker build -t opendigger-jupyter-python:1.0 ${pwd}` and then use`docker run -it --rm -p 8888:8888 -v ${pwd}:/python_kernel/notebook opendigger-jupyter-python:1.0` to build and run.

frank-zsy:
Thanks, I will try~

frank-zsy:
I am using the Python kernel and it mostly works fine, but there are some errors dealing with Neo4j database, I will fix them and sync back to the PR. @Best-Eminem

Best-Eminem:
> I am using the Python kernel and it mostly works fine, but there are some errors dealing with Neo4j database, I will fix them and sync back to the PR. @Best-Eminem

Ok, Thanks~

src/metrics/basic.py
```diff
@@ -0,0 +1,281 @@
...
+        # for every month individual, every element belongs to a individual element
+        forEveryMonthByConfig(config, lambda y, m: timeRangeSumClauseArray.append(['COALESCE({}_{}{}, 0.0)'.format(type, y, m)]))
+    elif config.get('groupTimeRange') == 'quarter':
+        # for every quarter, need to find out when to push a new element by quarter
+        lastQuarter = 0
+        forEveryMonthByConfig(config, process_quarter)
+    elif config.get('groupTimeRange') == 'year':
+        # for every year, need to find out when to push a new element by the year;
+        lastYear = 0
+        forEveryMonthByConfig(config, process_year)
```
frank-zsy, line 165: The `groupTimeRange` config is not properly handled now, right?

frank-zsy: I am not quite familiar with Python, but now with all kinds of `groupTimeRange`, the query will always be like by `month`.

Best-Eminem: > The `groupTimeRange` config is not properly handled now, right?

I don't think I quite understand what you mean.

frank-zsy: For Neo4j query with different `groupTimeRange` should generate different query to get the proper result. If the `groupTimeRange` is `year` then data in a year should be aggregate into a single element in the result.

But right now since in Python, a function can not access an external variable, so lastYear in `process_year` will always be 0 and `lastYear` update will not effect the external variable. So the data are not aggregated by year now.

I fix this by remove `lastYear` and `lastQuarter`, then use `month` to judge when to push a new element. I will update this change back too.

Best-Eminem: > 

I have already fix it, and I will update it later.

frank-zsy: `nonloacl` will fix this too.

But I am using

``` python
if m in [1, 4, 7, 10] or len(timeRangeSumClauseArray) == 0: timeRangeSumClauseArray.append([])
```

for `process_quarter` and 

``` python
if m == 1 or len(timeRangeSumClauseArray) == 0: timeRangeSumClauseArray.append([])
```

for `process_year`.

I think this will do too.

frank-zsy:
Another question:

- What does `src/metrics/__init.py__` file do? Is this file needed or any other `__init.py__` file needed? If no such file needed, can we add it into `.gitignore`?
- What do `*/**/__pychache/*` do? Can we disable these files from being generated? If not, can we add them into `.gitignore`?

Best-Eminem:
> Another question:
> 
> * What does `src/metrics/__init.py__` file do? Is this file needed or any other `__init.py__` file needed? If no such file needed, can we add it into `.gitignore`?
> * What do `*/**/__pychache/*` do? Can we disable these files from being generated? If not, can we add them into `.gitignore`?

I ignore `__pycache__` in my local workspace, but I didn't commit that. `__pycache__` is the python cache files which can be ignored, `__init.py__` also should be ignored, I will update `.gitignore` later.

src/metrics/basic.py
Best-Eminem: Yes, both of them will work.

frank-zsy:
Don't bother, I will update `.gitignore` and update all the changes into this PR. You don't need to add more commits now.

Best-Eminem:
> Don't bother, I will update `.gitignore` and update all the changes into this PR. You don't need to add more commits now.

Ok

src/db/clickhouse.py
```diff
@@ -0,0 +1,22 @@
...
+from matplotlib.pyplot import get
+from config import getConfig
+from clickhouse_driver import Client
+
+_client = None
+def getClient(): 
+    global _client
+    if _client == None: 
+        config = EasyDict((getConfig()).get('db').get('clickhouse'))
+        _client = Client(config.host, config.port, config.database, config.user, config.password)
```
frank-zsy, line 13: Need to pass in `user` and `password` to access online database

src/db/neo4j_driver.py
```diff
@@ -0,0 +1,18 @@
...
+def getClient():
+    global _driver
+    if _driver == None:
+        neo4j_config = EasyDict(EasyDict(getConfig()).db.neo4j)
+        _driver = Graph(neo4j_config.host)
+    return _driver
+      
+def query(query_sql):
+    result = getClient().run(query_sql) # return a cursor object
+    return result.data()  # transform cursor to list
```
frank-zsy, line 18: Use `result.data()` to transform cursor to result list, so upper layer code can use the result directly.

src/metrics/basic.py
```diff
@@ -0,0 +1,282 @@
...
+        for m in range(begin_month, end_month + 1):
+            func(y, m)
+
+# Repo
+def getRepoWhereClauseForNeo4j(config):
+    def process(l):
+        data = getGitHubData([l])
+        data = EasyDict(data)
+        arr = []
+        if len(data.githubRepos) > 0: arr.append('r.id IN {}'.format(data.githubRepos))
```
frank-zsy, line 59: All `array` or `list` param format need no string construction manually, Python can convert list to string that database can understand in SQL. So remove all `[]` and string construction.

frank-zsy:
I have updated this PR and add comments about what I change. And I ran all the `notebook_py.ipynb` codes and generate the results. Can you checkout the new code and double check this PR and the results in the notebook? @Best-Eminem

Best-Eminem:
> I have updated this PR and add comments about what I change. And I ran all the `notebook_py.ipynb` codes and generate the results. Can you checkout the new code and double check this PR and the results in the notebook? @Best-Eminem

Ok, I will check that.

src/metrics/activity_openrank.py
```diff
@@ -21,7 +21,7 @@
 REVIEW_COMMENT_WEIGHT = 4
 PULL_MERGED_WEIGHT = 2
 
-def getRepoActivityOrOpenrank(config, type='activity'):
+def getRepoActivityOrOpenrank(config, calType='activity'):
     """_summary_
 
     Args:
```
Best-Eminem, line 27: In annotation, the `type` need to be changed to `calType`.

Best-Eminem:
I find that some results in `handbook_py.ipynb` are different from results in `handbook_zh.ipynb`, is it because the `handbook_zh.ipynb` is not the latest edition?

src/metrics/activity_openrank.py
frank-zsy: Sure, I will fix this

frank-zsy:
> I find that some results in `handbook_py.ipynb` are different from results in `handbook_zh.ipynb`, is it because the `handbook_zh.ipynb` is not the latest edition? @frank-zsy

Yes, actually we changed `OpenRank` algorithm so the `activity` metric from Neo4j is not equal to `activity` from Clickhouse. This is what we should fix in the future that we can only get `activity` from Clickhouse and `OpenRank` from Neo4j.

Another reason is label data changed, we add a lot of new label data after the update of `handbook_zh.ipynb`.

Best-Eminem:
> Yes, actually we changed `OpenRank` algorithm so the `activity` metric from Neo4j is not equal to `activity` from Clickhouse. This is what we should fix in the future that we can only get `activity` from Clickhouse and `OpenRank` from Neo4j.
> 
> Another reason is label data changed, we add a lot of new label data after the update of `handbook_zh.ipynb`.

Ok, got it.

frank-zsy:
If there are no more problems, I will merge this PR tommorrow.

PR #967, opened 2022-09-13
[metrics] modify index interface of python kernel

### LINKED ISSUE #962: [Feature] Modify activity interface
In last version, `getActivity` will use Neo4j to get result and `getActivityWithDetail` will use Clickhouse.

But as we changed OpenRank algorithm, the activity scores in Neo4j are not equal to the result from Clickhouse now.

So we need to remove `getActivityWithDetail` function and change `getActivity` to use Clickhouse to get result.

Then Neo4j will only be used to get `OpenRank` result.

frank-zsy:

close #962 

Remove `getRepoActivityWithDetail` function and change `getRepoActivity` to use Clickhouse driver and `getRepoOpenrank` only use Neo4j driver.

### DIFF
--- a/src/metrics/basic.py
+++ b/src/metrics/basic.py
@@ -1,4 +1,5 @@
 from itertools import groupby
+import db.clickhouse as clickhouse
 from numpy import append
 from label_data_utils import getGitHubData, getLabelData
 import datetime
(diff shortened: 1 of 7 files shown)


src/metrics/basic.py
```diff
@@ -85,9 +86,17 @@ def process(l):
         return '({})'.format(' OR '.join(arr))
     repoWhereClauseArray = []
     if config.get('repoIds'): repoWhereClauseArray.append('repo_id IN {}'.format(config.get('repoIds')))
-    if config.get('repoNames'): repoWhereClauseArray.append('repo_name IN {}'.format(config.get('repoNames')))
+    if config.get('repoNames'):
```
frank-zsy, line 89: As all names for repo, org and user will be used in group function, so we need to get id first and pass into where clause rather than use names in where clause.

Issue #979, opened 2022-09-28
[Proposal] data_description.csv should be updated


zhicheng-ning:
Hi, guys! [data_description](https://github.com/X-lab2017/open-digger/blob/master/docs/assets/data_description.csv) should be updated. In our newest github_log.events table, there are `132` field name, but in old `data_description.csv`, there are `142` field name.

```jupyter
0                             id
1                           type
2                         action
3                       actor_id
4                    actor_login
                 ...            
127          commit_comment_path
128      commit_comment_position
129          commit_comment_line
130    commit_comment_created_at
131    commit_comment_updated_at
Name: name, Length: 132, dtype: object
```

xgdyp:
Thanks for your feedback!
Currently `github_log.events` table diff `data_descripition.csv` is:
Added:
```{'body', 'pull_merge_commit_sha'}```
Deleted:
```{'release_assets', 'issue_comment_body', 'pull_merged_commit_sha', 'gollum_pages', 'issue_body', 'created_date', 'issue_assignees', 'push_commits', 'pull_review_comment_body', 'issue_labels',  'commit_comment_body', 'push_before'}```

Why these fields deleted except some xx_body fields ? Can I update the data_description.csv to currently version?
@frank-zsy

frank-zsy:
@xgdyp Yes, you can update the file of course. The update is done like several month ago to reduce the storage of the data(https://github.com/X-lab2017/open-research/issues/73). I think there are also no `assignee` and `label` data in the original log files.

xgdyp:
why pull_merged_commit_sha changed to pull_merge_commit_sha? Is there other reason?

frank-zsy:
> why pull_merged_commit_sha changed to pull_merge_commit_sha? Is there other reason?

I think it is a typo in the description file, the field is `pull_merge_commit_sha` in the first edition of the table schema.

PR #982, opened 2022-09-30
fix: update data_description.csv

xgdyp:
related to #979


xgdyp:
why `pull_merged_commit_sha` changed to `pull_merge_commit_sha`?

PR #994, opened 2022-10-11
Correct the command in start_server.sh with the same reason of #993.

yoyo-wu98:
Like #993 , we also need to correct this command.


xgdyp:
close #993

PR #1000, opened 2022-10-14
ci: auto mark issue reply status

### LINKED ISSUE #999: [CI] auto add issues comment reply status
Open-digger may also need to have this workflow according to what @xgdyp  said. 
The workflow reference to https://github.com/hypertrons/hypertrons-crx/issues/475

zhicheng-ning:
close #999

### DIFF
--- a/.github/workflows/issue-open.yml
+++ b/.github/workflows/issue-open.yml
@@ -1,0 +1,18 @@
+# issue-open.yml
+# mark the initial status of the issue when it's been opened
+name: Issue Open Check
+
+on:
+  issues:
+    types: [ opened ]
+
+jobs:
+  issue-open-add-labels:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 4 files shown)


.github/workflows/issue-open.yml
```diff
@@ -0,0 +1,18 @@
+# issue-open.yml
+# mark the initial status of the issue when it's been opened
+name: Issue Open Check
```
xgdyp, line 3: Can the two workflow merged into one? Or we rename the files like `reply-lable-issue-comment.yml` and `reply-lable-issue-open.yml`, Is this better?

zhicheng-ning: In a nutshell, the trigger time of two workflow is different, so we can't merged them into one workflow. As for rename the file, I think it's depends on whether it has clean meaning.

Issue #579, opened 2021-06-02
[Metrics] Code Changes Lines


xiaoya-yaya:
**Description**
- The sum of the number of lines touched (lines added plus lines removed) in all changes to the source code during a certain period
- Lines of code will be any line of a source code file, including comments and blank lines.

**Objective**
Although code changes can be a proxy to the coding activity of a project, not all changes are the same. Considering the aggregated number of lines touched in all changes gives a complementary idea of how large the changes are, and in general, how large is **the volume of coding activity**.

**Implementation**
[...]

PR #1037, opened 2022-11-09
metric:code change lines

longyanz:
Add Code Change Lines metric from CHAOSS
#579


frank-zsy:
LGTM, although it looks like there is a little problem in `order` related part but it is a common problem for all the existing metrics, I will fix it after merge.

/approve

PR #1036, opened 2022-11-09
feat: new export method

frank-zsy:
Signed-off-by: frank-zsy <syzhao1988@126.com>

close #1033 

What I have done in this PR:

- I add a `repo_openrank` and `user_openrank` table in Clickhouse online server to query OpenRank for any repo and user on GitHub for any month.
- Change OpenRank query from Neo4j to Clickhouse.
- Move `activity`, `openrank` and `attention` into `indices` file.
- Remove `hacking_hero` and `hypercrx` related tasks, instead using `monthly_export` task to export all the data to OSS
  - The task will sample repo and user about 2 million each to export
  - The task will export all the metrics and indices implemented in OpenDigger
  - All the metrics and indices will be stored in different files

PR #1038, opened 2022-11-10
metric:new contributors

### LINKED ISSUE #592: [Metrics] New Contributors
This is a specific implementation of the [Contributor](https://github.com/X-lab2017/open-digger/issues/577) #577  metric.

For each Contributor, only consider their first contribution.

image

**Resource**
CHAOSS: Evolution-> Community Growth
https://chaoss.community/metric-new-contributors/

CHAOSS: Common Metrics -> Who
https://chaoss.community/metric-contributors/

longyanz:
Add New Contributors metric from CHAOSS
close #592


src/metrics/chaoss.ts
```diff
@@ -411,4 +411,54 @@ FROM
...
+        count() as count
+        FROM(
+         SELECT min(created_at)as created_at,org_id,repo_id,name
+         FROM(
+          SELECT created_at,org_id,repo_id,arrayJoin(push_commits.name) as name
+          FROM github_log.events
+          WHERE ${whereClauses.join(' AND ')}
+           )AS t
+         GROUP BY org_id,repo_id,name
+      )
```
frank-zsy, line 444: - Please make sure that any SQL keyword should be in uppercase, like `COUNT`, `MIN`, `AS`.
- Please make sure to leave a blank space between columns, like `MIN(created_at) AS created_at, org_id, repo_id`, and etc.
- Actually I think `MIN(created_at)` and `arrayJoin` can be in a single SQL like:
```sql
SELECT MIN(created_at) AS created_at, org_id, repo_id, arrayJoin(push_commits.name) AS name
...
GROUP BY org_id, repo_id, name
```

src/metrics/chaoss.ts
```diff
@@ -411,4 +411,54 @@ FROM
...
-  };
\ No newline at end of file
+  };
+
+  export const chaossNewContributors = async (config: QueryConfig) => {
+    config = getMergedConfig(config);
+    const whereClauses: string[] = ["type = 'PushEvent'"];
+    const repoWhereClause = getRepoWhereClauseForClickhouse(config);
+    if (repoWhereClause) whereClauses.push(repoWhereClause);
+    // whereClauses.push(getTimeRangeWhereClauseForClickhouse(config));
```
frank-zsy, line 421: You can delete this line if you don't need the time range clause here.

src/metrics/chaoss.ts
```diff
@@ -411,4 +411,54 @@ FROM
...
+    
+    const result: any = await clickhouse.query(sql);
+    return result.map(row => {
+    const [ id, name, total_count, count ] = row;
+    return {
+      id,
+      name,
+      total_count,
+      count,
+      ratio: count.map(v => `${(v*100/total_count).toPrecision(2)}%`),
```
frank-zsy, line 461: In the CHAOSS metrics description: https://chaoss.community/metric-new-contributors/ , we don't really need a `ratio` filed in the result. Instead, we should return a `detail` field with all the new contributors name and how many commits did they make in the time period.

longyanz:
Ok， I will fix it.

frank-zsy:
@longyanz You can fix them in this PR, you don't need to close the PR, just add more commits to this PR.

frank-zsy:
For more information, I think we can also add another implementation using merged PR to find contributors since for most open source project, PR is used as the only contributing method.

And like in OpenDigger, I used Frankzhaopku and frank-zsy as my commit user name, so I will be new contributor twice using commit info.

longyanz:
> For more information, I think we can also add another implementation using merged PR to find contributors since for most open source project, PR is used as the only contributing method.
> 
> And like in OpenDigger, I used `Frankzhaopku` and `frank-zsy` as my commit user name, so I will be new contributor twice using commit info.

I discussed this problem with @xgdyp before，we will consider how to better implement this metric

frank-zsy:
@longyanz How is the process here? I will merge #1036 after this metric is merged.

longyanz:
> @longyanz How is the process here? I will merge #1036 after this metric is merged.

We're trying to finish it by this Wednesday.

frank-zsy:
So can I merge #1036 first, you may need to handle a little bit conflicts.

longyanz:
> So can I merge #1036 first, you may need to handle a little bit conflicts.

Ok, thanks!

xgdyp:
Hello @frank-zsy  @longyanz , can we implement this metric by counting PR creators whose PR merged?
There are three steps:
- write a function to count PR contributors from a specific period time.
the sql like :
```
select * from github_log.events where repo_name = 'X-lab2017/open-digger' and type= 'PullRequestEvent' and action = 'opened' and issue_id in (select issue_id from github_log.events where type = 'PullRequestEvent' and pull_merged = 1 and action = 'closed' and repo_name = 'X-lab2017/open-digger'and created_at > toDate('2022-10-1')  and created_at < toDate('2022-10-15'))
```
- calculate the contributors from `2015-1-1` to a specific start time
- calculate the contributors from start time to end time 
- get the difference from the two returned list

frank-zsy:
> Hello @frank-zsy @longyanz , can we implement this metric by counting PR creators whose PR merged? There are three steps:
> 
> * write a function to count PR contributors from a specific period time.
>   the sql like :
> 
> ```
> select * from github_log.events where repo_name = 'X-lab2017/open-digger' and type= 'PullRequestEvent' and action = 'opened' and issue_id in (select issue_id from github_log.events where type = 'PullRequestEvent' and pull_merged = 1 and action = 'closed' and repo_name = 'X-lab2017/open-digger'and created_at > toDate('2022-10-1')  and created_at < toDate('2022-10-15'))
> ```
> 
> * calculate the contributors from `2015-1-1` to a specific start time
> * calculate the contributors from start time to end time
> * get the difference from the two returned list

The implementation here is not quite right, you should understand the underlying functions of OpenDigger, actually the implementation should be just like the one in this PR but using PR author rather than commit author as group unit.

If we want to implement the metric by PR, we can add a `by` option to `options` like other metrics and using commit author by default.

xgdyp:
> The implementation here is not quite right, you should understand the underlying functions of OpenDigger, actually, the implementation should be just like the one in this PR but using PR author rather than commit author as a group unit.
> 
> If we want to implement the metric by PR, we can add a `by` option to `options` like other metrics and use commit author by default.

But the `PR author ` may be hard to get because in PushEvent's log, `actor_login` is not equal to a PR author.
An example like this:
```
select created_at,push_commits.name,actor_login, `push_commits.message`,push_distinct_size,pull_merge_commit_sha,* from github_log.events where repo_id ='288431943' and type = 'PushEvent' and created_at > toDate('2022-10-1')  and created_at < toDate('2022-12-1') order by created_at desc
```
image

frank-zsy:
There are `issue_author_id` and `issue_author_login` columns which can be used to find out the pull request author when a pull request is merged, just like how `activity` index was implemented.

xgdyp:
> There are `issue_author_id` and `issue_author_login` columns which can be used to find out the pull request author when a pull request is merged, just like how `activity` index was implemented.

I tried but in PushEvent, these columns are all none.
```
select created_at,push_commits.name,actor_login, issue_author_login,issue_author_id,issue_number,`push_commits.message`,push_distinct_size,pull_merge_commit_sha,* from github_log.events where repo_id ='288431943' and type = 'PushEvent' and created_at > toDate('2022-10-1')  and created_at < toDate('2022-12-1') order by created_at desc
```
image

I'm wondering why they are none too.

frank-zsy:
@xgdyp You should use `PullRequestEvent` like this https://github.com/X-lab2017/open-digger/blob/master/src/metrics/indices.ts#L97

xgdyp:
> @xgdyp You should use `PullRequestEvent` like this https://github.com/X-lab2017/open-digger/blob/master/src/metrics/indices.ts#L97
OK. I'll read this first.

xgdyp:
> @xgdyp You should use `PullRequestEvent` like this https://github.com/X-lab2017/open-digger/blob/master/src/metrics/indices.ts#L97

I know what you mean. I forget clickhouse has argmin function😅

xgdyp:
hi @frank-zsy , you can review it now~

src/metrics/chaoss.ts
```diff
@@ -539,3 +540,95 @@ FORMAT JSONCompact`;
...
+          WHERE ${whereClauses.join(' AND ')}
+          ${(config.options?.withBot && by !== 'commit') ? '' : "HAVING author NOT LIKE '%[bot]'"}
+        )
+      GROUP BY repo_id, ${by === 'commit' ? 'author' : 'actor_id' }
+      HAVING first_time >= toDate('${config.startYear}-${config.startMonth}-1') AND first_time < toDate('${endDate.getFullYear()}-${endDate.getMonth() + 1}-1')
+    )
+    GROUP BY id, time
+  )
+  GROUP BY id
+  ORDER BY new_contributors[-1] ${config.order}
```
frank-zsy, line 620: You can look into the metrics implemented right now, there maybe query without `order` option. Only order if the `order` option is set.

src/metrics/chaoss.ts
```diff
@@ -539,3 +540,95 @@ FORMAT JSONCompact`;
...
+            })()},
+            created_at
+          FROM github_log.events
+          WHERE ${whereClauses.join(' AND ')}
+          ${(config.options?.withBot && by !== 'commit') ? '' : "HAVING author NOT LIKE '%[bot]'"}
+        )
+      GROUP BY repo_id, ${by === 'commit' ? 'author' : 'actor_id' }
+      HAVING first_time >= toDate('${config.startYear}-${config.startMonth}-1') AND first_time < toDate('${endDate.getFullYear()}-${endDate.getMonth() + 1}-1')
+    )
+    GROUP BY id, time
```
frank-zsy, line 617: If `limit` option is set, you should also add a `LIMIT` clause here to avoid more data returned. I think I got `bus_factor` implemented wrong, actually we don't need a `LIMIT` clause outside but only here.

Or maybe we should discuss the logic of how should we implement limit option.

frank-zsy:
@xgdyp Yes, if a commit is co-authored by multiple users, we can not really know how much contribution comes from each author, so we can only count it to a certain random author.

frank-zsy:
@longyanz Since I merged #1052 , there are conflicts in notebooks, you may need to add the section again.

src/metrics/chaoss.ts
longyanz: Hi，do you mean we should not add `ORDER BY new_contributors[-1] ${config.order}` if the `order` option isn't set . If so，should we change the function `getMergedConfig` in `basic.ts`  ?
image

src/metrics/chaoss.ts
longyanz: I think add `limit` outside and add it here means two different implementation logics.

For example,if we use `labelIntersect: ['Company'], limit: 10` ,the first one is to find 10 repos with the most new contributors among all repos, and the second one is to randomly select 10 repos and sort them in descending order.

Is my understanding correct？Which logic do you think is better？

src/metrics/chaoss.ts
frank-zsy: If you `rebase` to branch to current master branch, I have removed the `order` field in `defaultConfig`.

src/metrics/chaoss.ts
frank-zsy: Not really.

`LIMIT` outside means calculate the metric for all repos and all time period and return top results ordered by the last value.

`LIMIT` inside should also contained with `ORDER` clause which means find top results in each time period and return  the list ordered by the last value.

The first one is better because it will calculate all the value but it is really resource consuming, the second is more economic but some value maybe lost.

frank-zsy: The first one returns all the data ordered by last value, so any top-10 item may not be included in the result since it is not one of the top 10 items in the last time period.

The second one will return all top 10 items for all time period which makes sense for applications like OpenLeaderboard.

src/metrics/chaoss.ts
longyanz: Sorry, I did not notice this. I will fix it.

src/metrics/chaoss.ts
longyanz: If we `limit` outside, we can order by their total new contributors and limit. This can avoid some situations like a repo has a lot contributors before but none contributors in the last period. We can implement this quickly.
image

If `limit` inside, what we do are `In a specific time period, which companies have the most new contributors. Is this right?
We will try to implement this tonight.

frank-zsy: Yes, I think we should use outside limit, that makes sense a lot.

longyanz:
Hi， @frank-zsy we implement the metric，can you review it？

frank-zsy:
LTGM. Thanks.

/approve

PR #1049, opened 2022-11-18
fix: precision and names in query option

### LINKED ISSUE #1046: [Bug] Precision config does not take effect right now.
Right now, for metrics which return float value, precision config does not take effect, like `activity`, `openrank`, `resolution_duration` and etc.

### LINKED ISSUE #1048: [Bug] Repo name, org name and user name option in metric query cause error.
Right now, in all metrics implementations, we use `repo_id`, `org_id` and `actor_id` as group columns rather than the names.

So we can not directly use names in where clause which will cause a SQL error, so we need to convert names to ids before the actual query.

frank-zsy:

fix #1046 
fix #1048 

- Fix precision spell
- Fix precision in query, add `ROUND()` to `getGroupArrayInsertAtClauseForClickhouse` function. Float, double and integer all fit with the `ROUND` function, but not for other types like array or complex data types. If some field is array type or complex type, use `noPrecision` in option to avoid it.
- Fix names in queries, convert them to `ids` before real query.


xgdyp:
Can I review it now?

frank-zsy:
@xgdyp Yes, it is ready for review.

PR #1063, opened 2022-11-23
fix metric: issue resolution duration

### LINKED ISSUE #1026: [bug] questions in Mertic:issue duration resolution
Hi @frank-zsy ,when I read `duration resolution ` metric in `handbook.ipynb`, I found this metric return a value about 4700 days. I think this maybe wrong
image

So I find some issues in our clickhouse and there `issue open` event missed. I think this caused the wrong result.
image
(https://github.com/microsoft/vscode/issues/122346)
and in this clause, `open_at` field return 1970-1-1 .
image

longyanz:
close #1026 
use `argMax(issue_created_at,created_at) AS opened_at `  instead of `maxIf(created_at, action = 'opened') AS opened_at` to avoid return `opened_at : '1970-1-1'`


xgdyp (commented):
I think these line charts may be wrong if using LIMIT

frank-zsy:
@xgdyp @longyanz The reason that the result is not quite make sense because the `issueResolutionDuration` metric uses `inside limit` clause like we discussed [here](https://github.com/X-lab2017/open-digger/pull/1038#discussion_r1026084184), I think we may need to handle this later, maybe we should add an option like `orderOption` which is `inside limit` or `outside limit`, we should use `outside limit` by default, but we really need `inside limit` for OpenLeaderboard.

frank-zsy:
I think we can merge this first and handle the `LIMIT` problem later.

/approve

PR #1067, opened 2022-11-25
refactor: add limitOption to metrics query

### LINKED ISSUE #1064: [Refactor] Optimize the `LIMIT` option in metrics interface.
We need to discuss the `LIMIT` option implementation in metrics.

Like we discussed in [here](https://github.com/X-lab2017/open-digger/pull/1038#discussion_r1026084184), for `LIMIT` option in metrics query, we have two different implementation logics.

- The first one is `inside limit` which means limit the result count for each time period.
- The second one is `outside limit` which means limit the result in the final step and need to calculate all the result first.

For example, like we need query a metric for companies during 2017 to 2019 and group by year and limit is 5, the first approach means get top 5 companies for each year and combine the result in final step. And the second approach means calculate the metric for all companies in all time period and sort the result by the latest value and then limit the count.

For common situation, the second approach is better and really make sense.

But for ranking list like OpenLeaderboard, we need the first approach since we really need the top 5 results for every time period.

Do I make this clear?

And I think we can add a new option `limitOption` and default is `outside`, and can be `inside`, for data export procedure like OpenLeaderboard, we can set it to `inside`.

frank-zsy:

close #1064 

- Add a new option in metric query options: `limitOption`, which is `each` or `all`, and set to `all` by default.
  - When using `each`, `LIMIT` clause will limit the result for every time period, which is a better solution for OpenLeaderboard
  - When using `all`, `LIMIT` clause will limit the result for the final result but not for each time period, which is a better way for normal query. So `all` is the default value.
- Set all the `limitOption` options to `each` for OpenLeaderboard task.
- Add the option into `handbook` notebook.


frank-zsy:
And @longyanz , when you add examples into notebook, you should always set `limit` to a positive number, I find you set `limit: -1` for `new contributors` metric, and you use `Company` label and group by repos, so the query will return thousands of rows which is not a good practice for example.

Even use `Company` as group unit, you also need to set the `limit` to 5 or 10 I think.

longyanz:
> 

Ok，I will ~

src/metrics/chaoss.ts
```diff
@@ -679,10 +701,13 @@ export const chaossNewContributors = async (config: QueryConfig<NewContributorsO
...
     )
     GROUP BY id, time
+    ${config.limitOption === 'each' && config.limit > 0 ? 
+    `${config.order ? `ORDER BY new_contributor ${config.order}` : ''} LIMIT ${config.limit} BY time` :
+    ''}
   )
   GROUP BY id
-  ${config.order ? `ORDER BY total_new_contributors ${config.order}` : ''}
-  ${config.limit > 0 ? `LIMIT ${config.limit}` : ''}
+  ${config.order ? `ORDER BY new_contributors[-1] ${config.order}` : ''}
```
xgdyp, line 709: Why this `ORDER BY` Clause use `new_contributors[-1]` instead of `SUM(new_contributors)`? Is it just to be consistent with other metrics?

frank-zsy: Yes, the logic should be identical for all metrics. The reason using `SUM(new_contriubtors)` is that we have no identical strategy about how to sort and limit the result.

frank-zsy: @xgdyp Is there any more problem? I think we can also add a `orderOption` later, but for now, identical logic is fine.

xgdyp (approved):

Issue #589, opened 2021-06-02
[Metrics] Issue Response Time


xiaoya-yaya:
**Description**
- This metric is an indication of how much time passes between the opening of an issue and a response from other contributors.
- This metric is a specific case of the **[Time to First Response metric](https://github.com/X-lab2017/open-digger/issues/574)** in the **Common working group.**

**Filters**
- response from role in project (e.g., first maintainer response)
- bot versus human (e.g., filter out automated “welcome first contributor messages”)
- opened by role in project (e.g., responsiveness to new contributors versus long-timers)
- date issue was opened
- status of issue (e.g., only look at currently open issues)

**Implementation**
- Average. Average response time in days.
- Median. Median response time in days.
![image](https://user-images.githubusercontent.com/56542972/120429849-aed44100-c3a8-11eb-8398-df095be7110f.png)

**Resource**
CHAOSS: Evolution -> Issue Resolution
https://chaoss.community/metric-issue-response-time/

longyanz:
Hi, @frank-zsy . Here is the logic to implement this metric. 

![1670219534783](https://user-images.githubusercontent.com/91717658/205559485-80a51ed5-434e-4c21-bf38-b3791a8698b1.png)

we use `issue_comments = 0` to avoid computing those transferred issues that already have comments. Is there anything that needs to be modified in this logic?

xgdyp:
> issue_comments = 0 
this can be interpreted as follows:
- If a issue create with `issue_comments` field not NULL, this issue must be a transferred issue otherwise it should be 0(If reopened, it should be reopened action)
- And if a Issue transferred from other repo and it has IssueComment already. We don't need to calculate this issue again because it has been responsed.
- If a Issue transferred from other repo and `IssueComments==0`. It means this Issue has not been response now so we calculate in current repo.

frank-zsy:
@longyanz There are several things I would like to know:
- When an issue is transferred from a repository to another, will there be a new `opened` action in event log?
- There is an implementation detail can be discussed, when an issue transferred to a new repository, should we calculate the response time again for the new repository? Since a transferred issue is a new issue for the new repository, why should the response in old repository be counted?
- For `responsed_at`, you use `created` action which is an action for comment related events, but I think `closed` is also a response right? Even label should be counted as a response but we have no label events in log.
[...]

Issue #586, opened 2021-06-02
[Metrics] Active Issues


xiaoya-yaya:
**Description**
- How many issues were active during a certain period?
- Issues showing some activity are those that had some comment, or some change in state (including closing the issue), during a certain period.
- In GitHub Issues, a comment, a new tag, or the action of closing an issue, is considered as a sign of activity.

**Implementation**
- Count. Total number of active issues during the period.
- Ratio. Ratio of active issues over total number of issues during that period.
- https://chaoss.biterg.io/app/kibana#/dashboard/GitHub-Issues

**Resource**
CHAOSS: Evolution -> Issue Resolution
https://chaoss.community/metric-issues-active/

xgdyp:
We will reach the metric by counting how many issues are active in a certain period?
for example:
input a start time : 2022.9.1 and a stop time:2022.10.1
then return a count.
Is this right?

xiaoya-yaya:
Yes. This metric measures the count of the number of issues that have been active during a given period.

PR #1086, opened 2022-12-06
metric：Issue Response Time

longyanz:
we can discuss the details here.

### DIFF
--- a/src/metrics/chaoss.ts
+++ b/src/metrics/chaoss.ts
@@ -306,29 +306,30 @@ FORMAT JSONCompact`;
 };
 
 interface IssueResponseTimeOptions {
-  type: 'avg' | 'median';
-  unit: 'week' | 'day' | 'hour' | 'minute';
+  thresholds: number[];
+  unit: 'month' | 'week' | 'day' | 'hour' | 'minute';
 }
 export const chaossIssueResponseTime = async (config: QueryConfig<IssueResponseTimeOptions>) => {
   config = getMergedConfig(config);
-  const whereClauses: string[] = ["type IN ('IssueCommentEvent','IssuesEvent') AND actor_login NOT LIKE '%[bot]'  "];
+  const whereClauses: string[] = ["type IN ('IssueCommentEvent', 'IssuesEvent') AND actor_login NOT LIKE '%[bot]'  "];
   const repoWhereClause = await getRepoWhereClauseForClickhouse(config);
   if (repoWhereClause) whereClauses.push(repoWhereClause);
   const endDate = new Date(`${config.endYear}-${config.endMonth}-1`);
   endDate.setMonth(config.endMonth);  // find next month  
-  let type = filterEnumType(config.options?.type, ['avg', 'median'], 'avg');
-  let unit = filterEnumType(config.options?.unit, ['week', 'day', 'hour', 'minute'], 'day');
+  let thresholds = config.options?.thresholds ?? [1, 3, 7]; // 3 thredholds will separate the value to 4 ranges
+  let ranges = [...thresholds, -1]; // an array with one more element than threholds
+  let unit = filterEnumType(config.options?.unit, ['month', 'week', 'day', 'hour', 'minute'], 'day');
 
   const sql = `
 SELECT 
  id,
  argMax(name, time) AS name,   
- ${getGroupArrayInsertAtClauseForClickhouse(config, { key: 'issue_response_time', defaultValue: 'NaN' })}
+ ${getGroupArrayInsertAtClauseForClickhouse(config, { key: 'issue_response_time', value: 'response_levels', defaultValue: `[${ranges.map(_ => `0`).join(',')}]`, noPrecision: true })}
 FROM
 (
   SELECT
-   ${getGroupTimeAndIdClauseForClickhouse(config, 'repo', 'issue_created_at')},
-   round(${type}(dateDiff('${unit}',issue_created_at, responsed_at)),${config.precision}) AS issue_response_time
+    ${getGroupTimeAndIdClauseForClickhouse(config, 'repo', 'issue_created_at')},
+    [${ranges.map((_t, i) => `countIf(response_level = ${i})`).join(',')}] AS response_levels
   FROM
   (
     SELECT
@@ -337,21 +338,24 @@ FROM
       org_id,
       argMax(org_login, created_at) AS org_login,
       issue_number,
-      minIf(created_at, action = 'opened' and issue_comments = 0  ) AS issue_created_at,
-      minIf(created_at, action = 'created' and actor_id != issue_author_id ) AS responsed_at
+      minIf(created_at, action = 'opened' AND issue_comments = 0) AS issue_created_at,
+      minIf(created_at, (action = 'created' AND actor_id != issue_author_id) OR (action = 'closed')) AS responded_at,
+      if(responded_at = toDate('1970-01-01'), now(), responded_at) AS first_responded_at,
+      dateDiff('${unit}', issue_created_at, first_responded_at) AS response_time,
+      multiIf(${thresholds.map((t, i) => `response_time <= ${t}, ${i}`)}, ${thresholds.length}) AS response_level
     FROM github_log.events
     WHERE ${whereClauses.join('AND')}
     GROUP BY repo_id, org_id, issue_number
     HAVING issue_created_at >= toDate('${config.startYear}-${config.startMonth}-1') 
-             AND issue_created_at < toDate('${endDate.getFullYear()}-${endDate.getMonth() + 1}-1') 
-              AND responsed_at != toDate('1970-1-1')
+             AND issue_created_at < toDate('${endDate.getFullYear()}-${endDate.getMonth() + 1}-1')
   )
-  GROUP BY id,time
-  ${config.order ? `ORDER BY issue_response_time ${config.order}` : ''}
-  ${config.limit > 0 ? `LIMIT ${config.limit} BY time` : ''}
+  GROUP BY id, time
+  ${config.limitOption === 'each' && config.limit > 0 ? 
+    `${config.order ? `ORDER BY response_levels[1] ${config.order}` : ''} LIMIT ${config.limit} BY time` :
+    ''}
 )
 GROUP BY id
-${config.order ? `ORDER BY issue_response_time[-1] ${config.order}` : ''}
+${config.order ? `ORDER BY issue_response_time[-1][1] ${config.order}` : ''}
 FORMAT JSONCompact`;
 
   const result: any = await clickhouse.query(sql);
--- a/src/metrics/chaoss.ts
+++ b/src/metrics/chaoss.ts
@@ -305,6 +305,66 @@ FORMAT JSONCompact`;
   });
 };
 
+interface IssueResponseTimeOptions {
+  type: 'avg' | 'median';
+  unit: 'week' | 'day' | 'hour' | 'minute';
+}
+export const chaossIssueResponseTime = async (config: QueryConfig<IssueResponseTimeOptions>) => {
+  config = getMergedConfig(config);
+  const whereClauses: string[] = ["type IN ('IssueCommentEvent','IssuesEvent') AND actor_login NOT LIKE '%[bot]'  "];
+  const repoWhereClause = await getRepoWhereClauseForClickhouse(config);
+  if (repoWhereClause) whereClauses.push(repoWhereClause);
+  const endDate = new Date(`${config.endYear}-${config.endMonth}-1`);
+  endDate.setMonth(config.endMonth);  // find next month  
+  let type = filterEnumType(config.options?.type, ['avg', 'median'], 'avg');
+  let unit = filterEnumType(config.options?.unit, ['week', 'day', 'hour', 'minute'], 'day');
+
+  const sql = `
+SELECT 
+ id,
+ argMax(name, time) AS name,   
+ ${getGroupArrayInsertAtClauseForClickhouse(config, { key: 'issue_response_time', defaultValue: 'NaN' })}
+FROM
+(
+  SELECT
+   ${getGroupTimeAndIdClauseForClickhouse(config, 'repo', 'issue_created_at')},
+   round(${type}(dateDiff('${unit}',issue_created_at, responsed_at)),${config.precision}) AS issue_response_time
+  FROM
+  (
+    SELECT
+      repo_id,
+      argMax(repo_name, created_at) AS repo_name,
+      org_id,
+      argMax(org_login, created_at) AS org_login,
+      issue_number,
+      minIf(created_at, action = 'opened' and issue_comments = 0  ) AS issue_created_at,
+      minIf(created_at, action = 'created' and actor_id != issue_author_id ) AS responsed_at
+    FROM github_log.events
+    WHERE ${whereClauses.join('AND')}
+    GROUP BY repo_id, org_id, issue_number
+    HAVING issue_created_at >= toDate('${config.startYear}-${config.startMonth}-1') 
+             AND issue_created_at < toDate('${endDate.getFullYear()}-${endDate.getMonth() + 1}-1') 
+              AND responsed_at != toDate('1970-1-1')
+  )
+  GROUP BY id,time
+  ${config.order ? `ORDER BY issue_response_time ${config.order}` : ''}
+  ${config.limit > 0 ? `LIMIT ${config.limit} BY time` : ''}
+)
+GROUP BY id
+${config.order ? `ORDER BY issue_response_time[-1] ${config.order}` : ''}
+FORMAT JSONCompact`;
+
+  const result: any = await clickhouse.query(sql);
+  return result.map(row => {
+    const [id, name, issue_response_time] = row;
+    return {
+      id,
+      name,
+      issue_response_time,
+    }
+  });
+};
+
 // Evolution - Code Development Efficiency
 export const chaossChangeRequestsAccepted = async (config: QueryConfig) => {
   config = getMergedConfig(config);
--- a/src/metrics/index.ts
+++ b/src/metrics/index.ts
@@ -1,7 +1,7 @@
 import { getRepoOpenrank, getRepoActivity, getUserOpenrank, getUserActivity, getAttention } from './indices';
 import { chaossCodeChangeCommits, chaossBusFactor, chaossIssuesNew, chaossIssuesClosed, chaossChangeRequestsAccepted, 
 chaossChangeRequestsDeclined, chaossIssueResolutionDuration, chaossCodeChangeLines, chaossTechnicalFork, 
-chaossChangeRequests, chaossChangeRequestReviews,chaossNewContributors,chaossChangeRequestsDuration } from './chaoss';
+chaossChangeRequests, chaossChangeRequestReviews,chaossNewContributors,chaossChangeRequestsDuration, chaossIssueResponseTime } from './chaoss';
 import { repoStars, repoIssueComments, repoParticipants } from './metrics';
 import { getRelatedUsers } from './related_users';
 
@@ -21,6 +21,7 @@ module.exports = {
   chaossChangeRequestsAccepted: chaossChangeRequestsAccepted,
   chaossChangeRequestsDeclined: chaossChangeRequestsDeclined,
   chaossIssueResolutionDuration: chaossIssueResolutionDuration,
+  chaossIssueResponseTime: chaossIssueResponseTime,
   chaossCodeChangeLines: chaossCodeChangeLines,
   chaossTechnicalFork: chaossTechnicalFork,
   chaossChangeRequests: chaossChangeRequests,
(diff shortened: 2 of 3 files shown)


longyanz:
Hi @frank-zsy , I will answer the questions in #589  here.

- For "When an issue is transferred from a repository to another, will there be a new `opened` action in event log?"
  - Yes. For example ,this issue is transferred from kubernetes/apiserver to kubernetes/kubernetes. 
    https://github.com/kubernetes/kubernetes/issues/110349
    Here's its log in the kubernetes/kubernetes repo.
    


- For "There is an implementation detail can be discussed, when an issue transferred to a new repository, should we calculate the response time again for the new repository? Since a transferred issue is a new issue for the new repository, why should the response in old repository be counted?"
  - As @xgdyp  answered in #589 ，maybe we can open a new issue to discuss how to handle transferred issue.

- For "For `responsed_at`, you use `created` action which is an action for comment related events, but I think `closed` is also a response right? Even label should be counted as a response but we have no label events in log." 
   - I think it involves another metric #586 . Personally, I think response only considers CommentEvent.

frank-zsy:
There is no specific definition what is a response to an issue in CHAOSS, but if we consider active issues, it explicitly defines active as comments, labels, close action and etc. If we decide to use only comment as a response, we need to add it to docs.

And in CHAOSS, response is constrained to an action performed by another developer which should not be the author of the issue.

src/metrics/chaoss.ts
```diff
@@ -305,6 +305,66 @@ FORMAT JSONCompact`;
...
+  FROM
+  (
+    SELECT
+      repo_id,
+      argMax(repo_name, created_at) AS repo_name,
+      org_id,
+      argMax(org_login, created_at) AS org_login,
+      issue_number,
+      minIf(created_at, action = 'opened' and issue_comments = 0  ) AS issue_created_at,
+      minIf(created_at, action = 'created' and actor_id != issue_author_id ) AS responsed_at
```
longyanz, line 341: @frank-zsy we can use 
`minIf(created_at, action in ( 'created','closed') and actor_id != issue_author_id ) AS responsed_at` 
instead of  
`minIf(created_at, action = 'created' and actor_id != issue_author_id ) AS responsed_at` to include close event.

xgdyp:
As discussed in #589, what should be the response time for these issues with no response?

longyanz:
Hi, @frank-zsy .If we set `the current date(Suppose we query today,it should return 2022-12-7) - issue_created_at` as the response time of an unresponded issue , the average will grow to a very large number. I think it may lead to inaccurate results. What kind of response time do you think is reasonable?

longyanz:
> Hi, @frank-zsy .If we set `the current date(Suppose we query today,it should return 2022-12-7) - issue_created_at` as the response time of an unresponded issue , the average will grow to a very large number. I think it may lead to inaccurate results. What kind of response time do you think is reasonable?

For example, We chose `2018-1`-`2018-12` issues from kubernetes/kubernetes. If we ignore unresponsed issue, 
the result is `(1,46.06),(2,73.57),(3,121.49),(4,125.34),(5,194.64),(6,77.84),(7,125.04),(8,122.9),(9,145.39),(10,142.39),(11,148.11),(12,55.57) hour` .
If we use `the current date(2022-12-7) - issue_created_at`, 
the result is `(1,132.99),(2,361.25),(3,365.14),(4,1025.12),(5,893.18),(6,392.45),(7,305.83),(8,611.63),(9,224.54),(10,703.66),(11,479.01),(12,315.2) hour` .

frank-zsy:
In my opinion, if we chose `2018-1` to `2018-12` as the time period, so the last time should be `2018-12-31` but not `2022-12-07`, the `endTime` should be used but not now.

I think the real question from me as a community manager is to find out the issues with longest response time, in this situation, if all issues without response are filtered, then I will miss all the issues with real problems which is no one really responds them.

For `average` result for a repository, I also think it makes sense. If you use `issue response time` metric to inspect a repository, then all the issues without response surely should be counted, otherwise a dead repository will have 0 response time after its death since no issue will be responded. It does not make sense to me.

And for what you are concern, I think `median` is a good approach to reflect the `average`(Intuitively but not mathematically) status of the issues response time for this repository, which means at most time, how long will be an issue responded in this repository.

xgdyp:
> In my opinion, if we chose `2018-01` to `2018-12` as the time period, so the end time should be `2018-12-31` but not `2022-12-07`, the `endTime` should be used but not the real `now`.

In this case, it also has some problems. For an example, we chose `2018-1` to `2018-12` as the time period.
If an issue created in `2018-1-1`, this issue has a full year to response so its possible response time range is [1,365]days.
But if an issue created in `2018-12-20`, for this issue, its possible range is[1,10]days. It's unfair(because this issue may be responsed at `2019-1-1` but its upper bound is only 10 days).

xgdyp:
So we thought about another method. We can handle these unresponsed issues with a max period of x-axis time-period(like if we use month, we choose 1 month or we use year we choose 1year) response time .
For an example, if we choose `year` option for X and `hour` for Y , we will get a result of 1year(or 365 *24hour) for these unresponsed issue. 
If we choose `quarter` option for X and `day` for Y , we will get a result of 1 quarter(or 90 days) for these unresponsed issue. 
Also, we can handle issues with a long response time too. If an issue responsed over a X period, it will return the X period time.

frank-zsy:
> In this case, it also has some problems. For an example, we chose `2018-1` to `2018-12` as the time period. If an issue created in `2018-1-1`, this issue has a full year to response so its possible response time range is [1,365]days. But if an issue created in `2018-12-20`, for this issue, its possible range is[1,10]days. It's unfair(because this issue may be responsed at `2019-1-1` but its upper bound is only 10 days).

I can not really understand why this is unfair? If an issue was created on `2018-12-20` and has not been responded until `2018-12-31`, then looking back from `2018-12-31`, it is surely not responded for 10 days and looking back from now, the issue is surely responded at least 10 days after its creation. So what is point you think it is unfair and how the unfairness effects the result?

xgdyp:
Becuase our aim is to set unresponsed issues to  a related big value. But for an extreme example:
If we choose 2018-1 to 2018-12 as the time period.  Average responsed time for responsed issues is 5 day. But in 12-31, 1000 issues are created so the total Issue Responsed Time will come to nearly `0` .

frank-zsy:
OK, then the thing is we should try our best to find out the real response time rather than estimating with end time, like for issues created in 2018-12-31, we also try to find the first response in the future even it is in 2022 than the set time period, does it make sense to you?

In this method, the meaning of this metric is find out the first response time for the ***issues created during this time period***, no matter when the first response time occurs.

But still, if we choose the time period with an end time quite close to `now`, then still we need to handle the no-responded issues, like you suggest, I think we should set a minimal time span for no-responded issues to avoid this.

xgdyp:
> OK, then the thing is we should try our best to find out the real response time rather than estimating with end time, like for issues created in 2018-12-31, we also try to find the first response in the future even it is much more later than the set time period like in 2022, does it make sense to you?

Yes, I mean this too.

> But still, if we choose the time period with an end time quite close to `now`, then still we need to handle the no-response issues, like you suggest, I think we should set a minimal time span for no-response issues to avoid this.

I'll think more on this question .

xgdyp:
So, our strategy is:
- for every issue, we will search a time-span (the time-span define to a X-axis time period such as `a year`、`a month`、`a season`) 
- If they have been responsed under the time-span, we set reponse time range for this issue (we called response in sql) as ` response_at(the first issue comment or close event ) - issue_created_at`.
- If they have been responsed over the time-span, we set them to the time-span
- If `responsed_at` equals to `1970-1-1`, this means we don't find response events for this issue. So it belongs to `unresponsed issues`, we set their time range to a time-span, too.

In this strategy, the max response time is a time-span, so if maintainers want to focus on unresponsed issues, they just foucs on the issues whose value is  a time-span . Also, these issues responsed over a time-span need to be foucsed because it's too long.

How do you think about this?

longyanz:
> So, our strategy is:
> 
> * for every issue, we will search a time-span (the time-span define to a X-axis time period such as `a year`、`a month`、`a season`)
> * If they have been responsed under the time-span, we set reponse time range for this issue (we called response in sql) as ` response_at(the first issue comment or close event ) - issue_created_at`.
> * If they have been responsed over the time-span, we set them to the time-span
> * If `responsed_at` equals to `1970-1-1`, this means we don't find response events for this issue. So it belongs to `unresponsed issues`, we set their time range to a time-span, too.
> 
> In this strategy, the max response time is a time-span, so if maintainers want to focus on unresponsed issues, they just foucs on the issues whose value is a time-span . Also, these issues responsed over a time-span need to be foucsed because it's too long.
> 
> How do you think about this?

Hi, @xgdyp, @frank-zsy. I try to use this strategy to implement this metric.The result is below,
8d8f940f6a928d310ce78e5808c8643
db51703d3eabf9626eabed96025e9de

As we can see, Baidu touched the max time-span 744(31days, a X axis time period * 24 hours, Y time period). If we talk about `median` graph, we can say: at least half of issues in Baidu opensource projects are not responsed or responsed over 1 month. And issues in our companies not like this.

frank-zsy:
@xgdyp I am still quite not understand why the max response time is the time-span, and for time-span, I think you mean the `groupTimeRange` option in the query. So if an issue is responded in 2 month, should we still set it to 1 month?

If we search the whole time range to `now`, for no-response issue, if the time from its creation to now is 2 years, so if there is a first response in the future, the first response time for the issue will definitely over 2 years, right? So what's the problem to set its response time to 2 years?

frank-zsy:
Still, I can not really follow what is the point here?

> 1. Using this strategy, we can set the unresponsed issue to the max value so we can focus on them more.

For community manager, if you want to find out the issues which are not responded yet, you can just find them but not in this metric. I think there are 2 disadvantages using max time-span as the first response time

- The time-span is different with different query options, so the results with different options leads to different results for these issues, it is quite confusing.
- All metrics should reflect objective results of objective data, but max time-span is an implicit rule, we should definitely not introduce any implicit rule into metrics implementation, let alone it can not return identical results for all options.

> 2. If an issue are created in 2015 and  responsed now, it's too long for this issue if we calc this metric by `now - 2015-x-xx`,because these issues who are not responsed only return a time-span value, and I think value for responsed issue
>    should lower than unresponsed issues( they are not responsed yet!)

If an issue was created in 2015 and responded now, so the first response time definitely should be `now - 2015-x-xx`, why do you think it is too long and we shouldn't use the real value?

And why do you think the value for responded issue should lower than no-response issue? If an issue was created in 2018-01-01 and responded in 2018-03-01, so the first response time for this issue is 3 months, and another issue was created 1 month ago and not responded for now, so the first response time is 1 month if we use real time span. So the value for first one is larger than the second one, what is the problem? Why does this not make sense to you?

xgdyp:
If you think we should set all responsed issues with real response time, we can do this.

I just think the value for responded issue should lower than no-response issue.
In your example,  ` another issue was created 1 month ago` may be responsed in `2026-1-1`, but we set the value to one month.
Another example, two issue created at the same time `2018-1-1`, the first was responsed in `2022-1-1` so its response time is 2year. The second has not been responsed so the value is `X axis time `(`groupTimeRange` such as 1 month).
actually the first issue is responsed early than the second but its value is larger than the second, so I think it's unfair.

frank-zsy:
> In your example, ` another issue was created 1 month ago` may be responsed in `2026-1-1`, but we set the value to one month. 

Because we definitely can not be sure when the issue will be responded in the future, but what we definitely sure is the issue's first response is larger than 1 month since it was created 1 month ago and still get no response.

> Another example, two issue created at the same time `2018-1-1`, the first was responsed in `2022-1-1` so its response time is 2year. The second has not been responsed so the value is `X axis time `(`groupTimeRange` such as 1 month). actually the first issue is responsed early than the second but its value is larger than the second, so I think it's unfair.

Yes, of course it is unfair. So if we use `now` as the response time, the first issue's response time is 2 years and the second issue's response time is 3 years which perfectly makes sense. Why should we set the second one to 1 month?

xgdyp:
> In my opinion, if we chose `2018-01` to `2018-12` as the time period, so the end time should be `2018-12-31` but not `2022-12-07`, the `endTime` should be used but not the real `now`.
First, we follow this thought.
Second, we also dedicate to make `avg` result more sensible though it maybe meaningless. Using `time span` can make the result's `upper_bound` to `time span` so the `avg` result can make sense.

So do you think now, we should use `now` more than `time span`

longyanz:
> > In your example, ` another issue was created 1 month ago` may be responsed in `2026-1-1`, but we set the value to one month.
> 
> Because we definitely can not be sure when the issue will be responded in the future, but what we definitely sure is the issue's first response is larger than 1 month since it was created 1 month ago and still get no response.
> 
> > Another example, two issue created at the same time `2018-1-1`, the first was responsed in `2022-1-1` so its response time is 2year. The second has not been responsed so the value is `X axis time `(`groupTimeRange` such as 1 month). actually the first issue is responsed early than the second but its value is larger than the second, so I think it's unfair.
> 
> Yes, of course it is unfair. So if we use `now` as the response time, the first issue's response time is 2 years and the second issue's response time is 3 years which perfectly makes sense. Why should we set the second one to 1 month?

I think using `now` will result in a small number of unresponsive issues that greatly inflate the average. It may not reflect the original response time characteristics of most issues.

frank-zsy:
I am not quite sure about this, I think I would like to discuss this one in WeChat channel.

xiaoya-yaya:
Miss-clicked and made the draft PR ready for review. Apologies!

src/metrics/chaoss.ts
xiaoya-yaya: If I understand this correctly, we decide to ignore issues that has no comments and closed by the author him/herself. Is that right?

frank-zsy: > If I understand this correctly, we decide to ignore issues that has no comments and closed by the author him/herself. Is that right?

I think so, we should ignore the issues closed by author without comments.

birdflyi:
> > > In your example, ` another issue was created 1 month ago` may be responsed in `2026-1-1`, but we set the value to one month.
> > 
> > 
> > Because we definitely can not be sure when the issue will be responded in the future, but what we definitely sure is the issue's first response is larger than 1 month since it was created 1 month ago and still get no response.
> > > Another example, two issue created at the same time `2018-1-1`, the first was responsed in `2022-1-1` so its response time is 2year. The second has not been responsed so the value is `X axis time `(`groupTimeRange` such as 1 month). actually the first issue is responsed early than the second but its value is larger than the second, so I think it's unfair.
> > 
> > 
> > Yes, of course it is unfair. So if we use `now` as the response time, the first issue's response time is 2 years and the second issue's response time is 3 years which perfectly makes sense. Why should we set the second one to 1 month?
> 
> I think using `now` will result in a small number of unresponsive issues that greatly inflate the average. It may not reflect the original response time characteristics of most issues.

I'd like to consider the unresponsed(or never-responsed) issue as the same. Simplicity is beauty. We should not filter any issue for our purpose, unless anyone can understand the meaning on first glimps of the matric name. That will need a lot qualifier. If someone really want to know the stage of those never-responsed issues, there will be a new metric for this special group of issues. A never-responsed issue is also a issue not responsed yet. `issue response time` is measrued by time and only time, that is, a issue always can be measured if there is a pair of create time and measure time.

Is it fair by using `now`? In my opinion, this is not a problem about the measurement, while it may be unfair for your task for the message you can get from this metric is not enough. We can add ever-responsed state to complement these message for your task. And we will get another metric: `responsed issue response time` to solve this.

frank-zsy:
To summarize what we discussed, I think the key points in this metic implementation are:

- Use discrete response level rather than simple value, I think 0 day, 1 to 3 days, 4 - 7 days and more than 7 days are good. I think we can also add this to options so users can customize the result.
- For no-response issue, we can also use `now` as response time since it will fall into the correct range(at least for now), or we can just ignore them, I think include no-response issue is just fine under discrete response level.
- Issue closed by author without response should be considered. But I think we should not just filter them out, we can change the condition of response, comment by non-author and close event no matter from author or not can be considered as a response. @xiaoya-yaya 
- To order the results, we can use the less than 0 day issues ratio as the indicator rather a real response time value.

Does this make sense to you? @longyanz

longyanz:
> I can help you to modify the implementation if you allow me. @longyanz

Of course yes. Thanks!!!

src/metrics/chaoss.ts
```diff
@@ -305,6 +305,70 @@ FORMAT JSONCompact`;
...
+  unit: 'month' | 'week' | 'day' | 'hour' | 'minute';
+}
+export const chaossIssueResponseTime = async (config: QueryConfig<IssueResponseTimeOptions>) => {
+  config = getMergedConfig(config);
+  const whereClauses: string[] = ["type IN ('IssueCommentEvent', 'IssuesEvent') AND actor_login NOT LIKE '%[bot]'  "];
+  const repoWhereClause = await getRepoWhereClauseForClickhouse(config);
+  if (repoWhereClause) whereClauses.push(repoWhereClause);
+  const endDate = new Date(`${config.endYear}-${config.endMonth}-1`);
+  endDate.setMonth(config.endMonth);  // find next month  
+  let thresholds = config.options?.thresholds ?? [1, 3, 7]; // 3 thredholds will separate the value to 4 ranges
```
frank-zsy, line 319: Add `thresholds` and `unit` options, default is <1 day, 1-3 days, 3-7 days and > 7 days. Users can customize their own options like [12, 24, 36, 48] hours.

src/metrics/chaoss.ts
```diff
@@ -305,6 +305,70 @@ FORMAT JSONCompact`;
...
+      repo_id,
+      argMax(repo_name, created_at) AS repo_name,
+      org_id,
+      argMax(org_login, created_at) AS org_login,
+      issue_number,
+      minIf(created_at, action = 'opened' AND issue_comments = 0) AS issue_created_at,
+      minIf(created_at, (action = 'created' AND actor_id != issue_author_id) OR (action = 'closed')) AS responded_at,
+      if(responded_at = toDate('1970-01-01'), now(), responded_at) AS first_responded_at,
+      dateDiff('${unit}', issue_created_at, first_responded_at) AS response_time,
+      multiIf(${thresholds.map((t, i) => `response_time <= ${t}, ${i}`)}, ${thresholds.length}) AS response_level
```
frank-zsy, line 345: Using thresholds to calculate `response_level` for every issue directly

src/metrics/chaoss.ts
```diff
@@ -305,6 +305,70 @@ FORMAT JSONCompact`;
...
+  const sql = `
+SELECT 
+ id,
+ argMax(name, time) AS name,   
+ ${getGroupArrayInsertAtClauseForClickhouse(config, { key: 'issue_response_time', value: 'response_levels', defaultValue: `[${ranges.map(_ => `0`).join(',')}]`, noPrecision: true })}
+FROM
+(
+  SELECT
+    ${getGroupTimeAndIdClauseForClickhouse(config, 'repo', 'issue_created_at')},
+    [${ranges.map((_t, i) => `countIf(response_level = ${i})`).join(',')}] AS response_levels
```
frank-zsy, line 332: Put all `response_level` into an array as `response_levels`

src/metrics/chaoss.ts
```diff
@@ -305,6 +305,70 @@ FORMAT JSONCompact`;
...
+  FROM
+  (
+    SELECT
+      repo_id,
+      argMax(repo_name, created_at) AS repo_name,
+      org_id,
+      argMax(org_login, created_at) AS org_login,
+      issue_number,
+      minIf(created_at, action = 'opened' AND issue_comments = 0) AS issue_created_at,
+      minIf(created_at, (action = 'created' AND actor_id != issue_author_id) OR (action = 'closed')) AS responded_at,
```
frank-zsy, line 342: Comments from another developer or close events are considered as response.

src/metrics/chaoss.ts
```diff
@@ -305,6 +305,70 @@ FORMAT JSONCompact`;
...
+    HAVING issue_created_at >= toDate('${config.startYear}-${config.startMonth}-1') 
+             AND issue_created_at < toDate('${endDate.getFullYear()}-${endDate.getMonth() + 1}-1')
+  )
+  GROUP BY id, time
+  ${config.limitOption === 'each' && config.limit > 0 ? 
+    `${config.order ? `ORDER BY response_levels[1] ${config.order}` : ''} LIMIT ${config.limit} BY time` :
+    ''}
+)
+GROUP BY id
+${config.order ? `ORDER BY issue_response_time[-1][1] ${config.order}` : ''}
```
frank-zsy, line 358: Order by the shortest response time issues count, do not use ratio here because I found using ratio leads to a problem that a repo with only 1 issue but response in 1 day will be 100% shortest response time, so it will be first but what we really want is the repos always response quickly.

Although in this method, big community repository will have advantage even the ratio is not high but just with more issues as base number.

frank-zsy:
@longyanz @xiaoya-yaya @xgdyp @birdflyi I have modify the PR and tested it in my local env, I do not add examples into notebook since with ClickHouse server upgrade, the notebook will be modified later.

Here is the result from my local env:
image

longyanz:
> @longyanz @xiaoya-yaya @xgdyp @birdflyi I have modify the PR and tested it in my local env, can you review on this.
> 
> I do not add examples into notebook since with ClickHouse server upgrade, the notebook will be modified later.
> 
> Here is the result from my local env: image

I think the results are pretty good！

frank-zsy:
@xgdyp Can you review this one?

xgdyp (approved):

xgdyp:
I've reviewed it. I think it's great!

PR #1097, opened 2022-12-14
metric: Change request acceptance ratio

longyanz:
Hi, i'm trying to implement this metric #584. I think the logic to implement it is clear，which is to count the number of accepted pr and the number of declined pr. Instance of implementation given by CHAOSS is a stack histogram（see the figure below）. I additional calculate `ratio =  merged pr/all closed pr`.

But I met a problem，even though I had read the Plotly documentation, I couldn't figure out how to plot a stack histogram with the current data, maybe that requires us to format the data. Could anyone please give me some advice？

This figure is the data that i get now.


frank-zsy:
If you use the following code in plotly's code pen, you will get a diagram with all the data with first 5 elements.

``` JavaScript
var rawData = {
  name: 'PingCAP',
  accept_count: [ 44, 42, 75, 96, 76],
  decline_count: [9, 2, 7, 9, 5],
};

var trace1 = {
  x: [1, 2, 3, 4, 5],
  y: rawData.accept_count,
  name: 'Accept',
  type: 'bar'
};

var trace2 = {
  x: [1, 2, 3, 4, 5],
  y: rawData.decline_count,
  name: 'Declined',
  type: 'bar'
};

var trace3 = {
  x: [1, 2, 3, 4, 5],
  y: [83, 95, 91, 91, 94],
  name: 'Accept Ratio',
  type: 'line'
}

var data = [trace1, trace2, trace3];

var layout = {barmode: 'stack'};

Plotly.newPlot('myDiv', data, layout);
```

Which looks like this:

image

src/metrics/chaoss.ts
```diff
@@ -520,6 +520,53 @@ FORMAT JSONCompact`;
...
+    ''}
+)
+GROUP BY id
+${config.order ? `ORDER BY change_requests_accepted_ratio[-1] ${config.order}` : ''}
+${config.limitOption === 'all' && config.limit > 0 ? `LIMIT ${config.limit}` : ''}
+FORMAT JSONCompact`;
+
+  const result: any = await clickhouse.query(sql);
+  return result.map(row => {
+    const [id, name, ratio,acount,dcount] = row;
```
frank-zsy, line 559: You missed blank space before `account` and `dcount`.

And I think you should use `accept_count`, `declined_count` and `accept_ratio` rather than `acount`, `dcount` and `ratio`, since OpenDigger data will release to OSS, and we should try our best to make sure users can understand the data without documentation.

src/metrics/chaoss.ts
```diff
@@ -520,6 +520,53 @@ FORMAT JSONCompact`;
...
+  ${getGroupArrayInsertAtClauseForClickhouse(config, { key: 'change_requests_declined', value: 'dcount' })}
+
+FROM
+(
+  SELECT
+    ${getGroupTimeAndIdClauseForClickhouse(config, 'repo')},
+    COUNT() AS count,
+    countIf(pull_merged = 1) AS acount,
+    countIf(pull_merged = 0) AS dcount,
+    round(acount/count,2) AS ratio
```
frank-zsy, line 544: We don't need to `round` here, `getGroupArrayInsertArrayClauseForClickhouse` will round by precision config for you. And you should use `precision` rather than `2` if you want to `round` here.

longyanz:
Ok~ I will fix it.

longyanz:
Hi, @frank-zsy .I met a problem when I was trying to draw the stack bar. Here is my code , there is no result on the graph. 

40233045ccfb21db0c72ab4c50fa939

I use `console.log();` to make sure that `trace1,trace2,stackdata` all contain data.

longyanz:
And I have one more question，If I want to show multiple repos in one graph.  As in this diagram, If I want to stack the `decline_count` of each repo for that month on top of each bar , what should I do ?

frank-zsy:
I will try to do some experiments on this metric, since you don't rebase the changes before, do you need me to rebase this PR for you?

frank-zsy:
There are really some tricks to make this work, although OpenDigger does not focus on data visualization but only data process, I am very interested in this one and I think I made a quite desirable result after reading the plotly API document and lots of test.

image

longyanz:
> There are really some tricks to make this work, although OpenDigger does not focus on data visualization but only data process, I am very interested in this one and I think I made a quite desirable result after reading the plotly API document and lots of test.
> 
> image
> 
> I use returned data array length to calculate the bar width and offset for each row, and use `acount` as base of `dcount` to make them stack together.

I think it's a excellent result! Thanks!

frank-zsy:
I also add `ratio` into another y axis to make all the data fit into a single chart.

image

And you may need to rebase the current master branch to use new ClickHouse driver and instance.

src/metrics/chaoss.ts
```diff
@@ -511,6 +511,53 @@ ${config.order ? `ORDER BY resolution_duration[-1] ${config.order}` : ''}`;
...
+  WHERE ${whereClauses.join(' AND ')}
+  GROUP BY id, time
+  ${config.limitOption === 'each' && config.limit > 0 ? 
+    `${config.order ? `ORDER BY ratio ${config.order}` : ''} LIMIT ${config.limit} BY time` :
+    ''}
+)
+GROUP BY id
+${config.order ? `ORDER BY change_requests_accepted_ratio[-1] ${config.order}` : ''}
+${config.limitOption === 'all' && config.limit > 0 ? `LIMIT ${config.limit}` : ''}
+FORMAT JSONCompact`;
```
frank-zsy, line 546: Did you test the code? With new ClickHouse instance and client, `FORMAT JSONCompact` like this should cause an error.

longyanz: > Did you test the code? With new ClickHouse instance and client, `FORMAT JSONCompact` like this should cause an error.

Sorry ，I missed it.

longyanz: I have corrected it.

src/metrics/chaoss.ts
```diff
@@ -511,6 +511,52 @@ ${config.order ? `ORDER BY resolution_duration[-1] ${config.order}` : ''}`;
...
+  FROM gh_events
+  WHERE ${whereClauses.join(' AND ')}
+  GROUP BY id, time
+  ${config.limitOption === 'each' && config.limit > 0 ? 
+    `${config.order ? `ORDER BY ratio ${config.order}` : ''} LIMIT ${config.limit} BY time` :
+    ''}
+)
+GROUP BY id
+${config.order ? `ORDER BY change_requests_accepted_ratio[-1] ${config.order}` : ''}
+${config.limitOption === 'all' && config.limit > 0 ? `LIMIT ${config.limit}` : ''}
```
frank-zsy, line 545: There is a missing \` at the end of this line which leads to an unclosed string and will break all the file.

longyanz: I have added it.

frank-zsy:
@longyanz What sample data you use to generate the examples in handbook notebook? I use current online service to generate the data but it is entirely different from yours.

image

frank-zsy:
Although I can not get the same data, but I think the logic is fine, I will merge this PR first.

/approve

Issue #1087, opened 2022-12-07
[Refactor] Upgrade Clickhouse server to 22.8


frank-zsy:
As discussed in https://github.com/X-lab2017/open-wonderland/issues/222 , we will upgrade our Clickhouse cloud service server version from 20.8 to 22.8 recently. The data upload progress will be completed in 2 days.

Then we need to upgrade OpenDigger kernel to make sure OpenDigger is compatible with the server version.

The js kernel need to change underlying Clickhouse client library from [`@apla/clickhouse`](https://www.npmjs.com/package/@apla/clickhouse) to [`@clickhouse/client`](https://www.npmjs.com/package/@clickhouse/client) which is the official js client for Clickhouse and support 22.8 as the minimal version.

For client upgrade, we also need to upgrade Clickhouse sample data server image to 22.8 to make sure it is compatible with OpenDigger.

So the things we need to do includes:

- [x] Upgrade js kernel underlying library. #1098 
- [x] Upgrade python kernel underlying library. #1113 
- [x] Upgrade Clickhouse sample data server image version. #1099 

When all above tasks finished, we can entirely transform from Clickhouse 20.8 to 22.8 with more great features created by Clickhouse in past 2 years.

PR #1098, opened 2022-12-15
refactor: change underlying clickhouse driver

frank-zsy:

related to #1087 

Upgrade underlying TypeScript/JavaScript client for ClickHouse, update from 20.8 to 22.8.

All notebook has been tested and updated with new client and new server instance, all works fine now.

Upper interface of all metrics does not change at all.

CAUTIONS:

- The result format should be passed into driver as a second parameter rather than simply in the query. Default format is `JSONCompactEachRow` which is the common format used for all metrics. If you want to use the driver directly, you should understand how to set the return format.
- As we may use ClickHouse to store more open source data but not just GHArchive data. The database name changes from `github_log` to `opensource`, and database is set in the client driver on init, so we could only set table name in query SQL. And the tables' names are also changed as below:

| old | new |
|:--:|:--:|
| events | gh_events |
| export_repo | gh_export_repo |
| export_user | gh_export_user |
| repo_openrank | gh_repo_openrank |
| user_openrank | gh_user_openrank |

We may add more data about open source supply chain and artifact data to ClickHouse server later.


frank-zsy:
This PR is only used to upgrade the kernel scripts part, sample data upgrade will be submitted later.

src/db/clickhouse.ts
```diff
@@ -1,26 +1,25 @@
...
 
-export async function queryStream<T = any>(q: string, onRow: (row: T) => void): Promise<void> {
+export async function queryStream<T = any>(q: string, onRow: (row: T) => void, options: any = {}): Promise<void> {
   return new Promise(async resolve => {
-    const stream = (await getClient()).query(q);
-    stream.on('data', (row: T) => onRow(row));
-    stream.on('end', () => {
-      resolve();
-    });
+    const resultSet = await (await getClient()).query({ query: q, format: 'JSONCompactEachRow', ...options });
```
xgdyp, line 19: So in python kernel, what we need to change is table name ,db engine and `FORMAT JSONCompact` like this format, is this right?

frank-zsy: Nope, this is JavaScript kernel with JavaScript client, for Python kernel, you guys need to test it, if current Python kernel client is compatible with the new ClickHouse server version, you don't need to change anything. You only need to change Python code if the client is not compatible with new server.

Keep in mind that: JavaScript/TypeScript kernel code and Python kernel code are totally independent.

frank-zsy: But for sure, you need to change the database and table name for queries.

xgdyp: OK

xgdyp (approved):

frank-zsy:
@xgdyp You can approve this PR if you think it is fine. There is no need to get all reviewers to approve, OpenDigger only requires 1 reviewer indeed.

PR #1102, opened 2022-12-20
feat: 2022 open source report data section outline

xgdyp:
Hello, this PR contains content that needs to be analyzed using OpenDigger in open source report data section. 
The full outline we discussed in https://github.com/X-lab2017/open-wonderland/issues/226.
Can @frank-zsy help implement these metrics? Thanks a lot!


frank-zsy:
@xgdyp Is this a python kernel notebook? Can you change to a Node.js kernel notebook so I can implement it.

xgdyp:
> @xgdyp Is this a python kernel notebook? Can you change to a Node.js kernel notebook so I can implement it.

OK

xgdyp:
I've changed it.

xgdyp:
before 12.23, is this ok？



---- Replied Message ----
| From | Frank ***@***.***> |
| Date | 12/20/2022 21:30 |
| To | ***@***.***> |
| Cc | ***@***.***>***@***.***> |
| Subject | Re: [X-lab2017/open-digger] feat: 2022 open source report data section outline (PR #1102) |

OK, is there a deadline?

frank-zsy:
As 2022 data will be completed at least in 2023.01.01, if you need the data before 12.23, so 2022.12 data will not be included, why is that?

xgdyp:
> As 2022 data will be completed at least in 2023.01.01, if you need the data before 12.23, so 2022.12 data will not be included, why is that?

Because the timeline need use finish this before 2023.1.1 . Or do you think we can implement these first and redo these in 2023.1.1?

frank-zsy:
@xgdyp Of course, I have implemented all the content and you can check if it is OK. For every table, I use OpenRank as sorting value and add more metrics in OpenDigger to each repo or company, like activity, stars, participants, new_contributors, bus factor, issue response time and issue resolution duration. You can use these data to visualize the report.

And also some minor changes to code base, including:

- Fix some format issues.
- Use label identifier as id and label name as name column in label group function, so downstream procedures could use the ids to do more work like this report.
- Expose org_id and org_login in `new_contributors` metric which is bug only triggered when group by label.
- Remove redundant `chaoss` prefix in OpenDigger JavaScript API.

xgdyp:
OK

frank-zsy:
@xgdyp Are you going to add more commits to this PR or this is good to merge?

frank-zsy:
@xgdyp Next time, when you get more commits from master branch, you should use `git fetch/rebase` rather than `git fetch/merge` to make sure the commits order is right.

xgdyp:
I think it can merge now.

tyn1998:
Hi @frank-zsy,

For me, now I avoid using rebase merges in developing branches especially if I want to merge the main/master branch that is ahead of my working branch. The main reason is that a rebase merge also require a force push, which will make a PR conversation confusing. Here is an illustration:

image

While a normal merge doesn't bring this kind of issue:

image

> you should use git fetch/rebase rather than git fetch/merge to make sure the commits order is right.

What do you mean by "make sure the commits order is right"? Do you try to keep all commits in a developing branch ordered in a single line? If so, I'm a little bit confused. Since in my view, the last `/approve` will do a rebase merge and that merge can keep all commits in the master branch in a single ordered line.

frank-zsy:
@tyn1998 Yes, you are right, I just forgot that with `/approve` we use rebase and squash merge to merge the PR, so all the changes will be squashed into a single commit and rebase master branch on merge. So yes, it is irrelevant if we use `rebase` or `pull` to synchronize the master branch.

And add more commits with comments surely is better than force push on collaboration, I think I just more comfortable with commit amend and force push which is not a good habit, especially when we use rebase and squash merge for PR. Thanks for pointing this out.

Issue #1081, opened 2022-12-04
[Feature] network data for Hypercrx might be regenerated with latest formulas


tyn1998:
[Hypercrx](https://github.com/hypertrons/hypertrons-crx) presents network charts in both `Perceptor`s of repo pages and of profile pages. At the moment, there are 4 APIs backing these charts:

- repo
  - related repos: https://hypertrons-oss.x-lab.info/repo/hypertrons/hypertrons-crx.json
  - developer network in this repo: https://hypertrons-oss.x-lab.info/repo/hypertrons/hypertrons-crx_top.json
- developer
  - related developers: https://hypertrons-oss.x-lab.info/actor/tyn1998.json
  - repo network: https://hypertrons-oss.x-lab.info/actor/tyn1998_top.json

These data reflect status of repos and developers in the past **180 days** and they are regenerated each week (I suppose?). However, the formulas underpinning the data generation are really outdated, which makes `Activity` of nodes not consistent with the latest `Activity` we use in other charts.

Also, the data mentioned above are deployed in another oss sever, which in my view is not necessary. So I suggest that these data can be implemented here in OpenDigger with the latest computing methods.

We can talk about the schema of the json files here as well.

frank-zsy:
Of course, we can make the data generated from OpenDigger and we can decide how to do this.

First, I think we can use some standard graph format like GDF, here is a [graph file format overview paper](https://www.researchgate.net/publication/330185503_Overview_of_Standard_Graph_File_Formats), we can discuss which one to use.

frank-zsy:
I think we use the JSON format at first place because the data can directly consumed by ECharts graph component. I think CSV format is also universal. As a user, which one do your prefer? @tyn1998

tyn1998:
JSON is a first-class citizen in JavaScript, while people using Python to do data analysis might be in favour of csv.

Both format is ok for me since a little data parse is not a big deal.

I had a glance at [GDF csv format](https://gephi.org/users/supported-graph-formats/gdf-format/) in the paper, I think it's good.

frank-zsy:
OK, then we should discuss how the data are generated now. The first version use repo - developer relationship to generate the network. Repo - repo and developer - developer also use repo - developer relationship too but just aggregate.

I think the repo - developer network and repo - repo network work just fine with activity index as relationship, but for developer - developer network, some users are confused because they don't really know the other developers in their collaborating network, we can use issue level data to make relationship for developers which may make sense much more.

WDYT.

[...]

PR #1104, opened 2022-12-22
refactor: modify node size of networks

frank-zsy:

related to #1081 

Use `activity` to replace node size of developer and repo in repos' developer networks and developers' repo networks.


xgdyp:
Is this change used for limit nums of nodes in each graph?

frank-zsy:
No, as we discussed in #1081, this change is about to change repos' developer network and developers' repo network node size from their global OpeRank value to corresponding activity score which makes more sense.

PR #1106, opened 2022-12-23
feat: add handle data loss for 2021.10

### LINKED ISSUE #1040: [Feature] Handle data lose
As GHArchive was offline in 2021.10 for about half a month, the statistical data in 2021.10 are all about 50% of 2021.9 and 2021.11.

So when we upload data to OSS, we can do a little trick to fix the problem. I think we can use like `0.15 * Value_20218 + 0.35 * Value_20219 + 0.35 * Value_202211 + 0.15 * Value_202212` to estimate the actually value of 2021.10, and still we should keep a field `202110_original` to store the real value.

frank-zsy:

close #1040 

In export data process, handle data loss of 2021.10 for statistical metrics for monthly data since GHArchive has a [data service failure](https://github.com/igrigorik/gharchive.org/issues/232#issuecomment-678798777) in 2021.10 for about 2 weeks.

Use 2021.8 - 2021.12 data to estimate 2021.10 data and store the original data to `2021-10-raw` field.

### DIFF
--- a/src/cron/tasks/monthly_export.ts
+++ b/src/cron/tasks/monthly_export.ts
@@ -52,9 +52,9 @@ const task: Task = {
 
     // start to export data for all repos and actors
     // split the sql into 20 pieces to avoid memory issue
-    const getPartition = async (type: 'User' | 'Repo'): Promise<Array<{min: number, max: number}>> => {
+    const getPartition = async (type: 'User' | 'Repo'): Promise<Array<{ min: number, max: number }>> => {
       const quantileArr: number[] = [];
-      for (let i = 0.05; i <= 0.95; i += 0.05) { 
+      for (let i = 0.05; i <= 0.95; i += 0.05) {
         quantileArr.push(i);
       }
       const partitions: any[] = [];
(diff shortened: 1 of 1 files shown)


src/cron/tasks/monthly_export.ts
```diff
@@ -99,59 +101,68 @@ const task: Task = {
...
+          if (!disableDataLoss && option.groupTimeRange === 'month' && content['2021-10']) {
+            // handle data loss in 2021.10 only when generate data by month
+            content['2021-10-raw'] = content['2021-10'];
+            const arr = ['2021-08', '2021-09', '2021-11', '2021-12'].map(m => content[m]);
+            if (!arr[3]) arr[3] = 0;
+            if (!arr[2]) arr[2] = 0;
+            if (!arr[0]) arr[0] = arr[3];
+            if (!arr[1]) arr[1] = arr[2];
+            content['2021-10'] = Math.round([0.15, 0.35, 0.35, 0.15].map((f, i) => f * arr[i]).reduce((p, c) => p + c));
+          }
```
frank-zsy, line 113: The process will only take effect if 202110 data point exists and only for statistical metric when group by month.

The basic logic for the process is using `0.15 * Value_202108 + 0.35 * Value_202109 + 0.35 * Value_202111 + 0.15 * Value_202112`.

But since some repo may just occur after 2021.10, so replace with 202112 if 202108 not exists and replace with 202111 if 202109 not exists.

And for all the metrics, use integer format rather than float format because some metrics requires to be an integer.

tyn1998: Hi @frank-zsy, I think it can be good if you also put details in this review comment as a code comment so possible future matainers are able to understand the code detail with ease and quickly :-)

frank-zsy: Thanks, good point, I will add more code comment in the future.

PR #1112, opened 2022-12-26
data: add repository label data

PureNatural:
I got the top 10000 star repositories through github api.

20 undergraduates and 20 postgraduates were invited to label the repositories.

According to [this paper](https://arxiv.org/pdf/1707.02327.pdf), we divide the repository technology labels into five categories: Libraries and frameworks, Application software, Software tools, System software and Non software.

After five rounds of labeling by the students, the following data results are obtained through rule matching and the judgment of the person in charge.

Application software | 2319
Libraries and frameworks | 3025
Software tools | 1738
System software | 417
Non software | 2501

I submit our results based on the existing data label files of open-digger.

If there is something wrong, I will revise it.

Much thanks.


frank-zsy:
@PureNatural Thanks for the data, but I think the data should not be under `technology`, right now, `technology` label is used for certain tech area like cloud native or database, but these labels are not quite same. They are more like another type system but I am not sure what name to use yet. Can you think of a new one instead of `technology`?

PureNatural:
@frank-zsy Thanks for you reply. 

According to [this paper](https://arxiv.org/pdf/1707.02327.pdf), the data should be under _application domain_

image

I think we can use it.

frank-zsy:
Yes, `application_domain` is great and if there will be any sub domains in the future?

PureNatural:
Yes, our next work will focus on sub domains.

frank-zsy:
That's cool, thanks for the data.

/approve

frank-zsy:
@PureNatural Sorry that I did realize the `type` should also be changed, can you change all `Tech-0` to `Domain-0`, and `Domain-1`, `Domain-2` should be used as sub domain type.

PureNatural:
Have done in this PR https://github.com/X-lab2017/open-digger/pull/1117.

PR #1113, opened 2022-12-27
feat: update clickhouse schema

xgdyp:
In this PR, I changed table name in python kernel.
I test all metrics excludes some about openrank in neo4j and it works well.

### DIFF
--- a/python/config.py
+++ b/python/config.py
@@ -13,7 +13,7 @@ config = {
       'password': '',
       'protocol': 'http:',
       'format': 'JSON',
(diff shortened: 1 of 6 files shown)


frank-zsy:
Thanks, actually the OpenRank implementation in Node.js kernel now also depends on ClickHouse database rather than Neo4j database. But I think Python kernel is out of date now, we can update Python kernel about metrics and indices in the future.

/approve

xgdyp:
OK, I'll update metric index recently.

PR #1127, opened 2023-01-04
ci: add test to opendigger

### LINKED ISSUE #1126: [CI] We should add simple tests to OpenDigger
Right now, OpenDigger does not have any test code to make sure all the functions are working as expected, like labels, metrics and indices.

But recently, with more and more functions added into OpenDigger, and some refactors also break the code. This implies that human review is not sufficient anymore for OpenDigger.

So I think we should add some simple tests to OpenDigger to make sure:

- All label data works fine and returns desirable results.
- All metrics can run successfully and return correct format data.
- All database drivers works fine.

I think a simple test structure with [mocha](https://www.npmjs.com/package/mocha) is enough for now.

frank-zsy:

Add a simple test structure for Node.js kernel in OpenDigger

close #1126 

- Use mocha as our test framework
- Add 3 test files for 
  - Database driver test, only for ClickHouse right now
  - Label data test
  - Metrics test

Also find some bugs in current code
- The `getGroupArrayInsertAtClauseForClickhouse` function should always set the default value and array length, which is not supported in old version ClickHouse.
- The `chaossIssueResponseTime` and `chaossChangeRequestsDuration` metrics lost the tailing `LIMIT` clause.


frank-zsy:
To solve the database connection in CI

- I created a sample dataset only with CI data(google org data in 2015 and 2016), with `sample_data/sql_files/ci_test.sql` file.
- Run a public ClickHouse instance for CI test and inject into `npm test` script.
- Remove OpenRank related test since it requires more data than original table.
- Extend a test case timeout to 15s since it will be slow to get data from ClickHouse in GitHub actions due to network.

So now, tests can be run successfully if the ClickHouse CI server instance is up.

frank-zsy:
Developers can use basic metrics test cases to test their metrics implementation, the basic test only make sure that:

- The metrics implementation should return data rather than raise any error.
- The metrics implementation should work fine with limit function.
- The metrics implementation should return correct array length data for its main field.

And developers can add more detailed tests to specific metrics too to test its own logic.

tyn1998:
> Still, PR from fork can not access the secret in actions, so we can not pass the online ClickHouse service into the CI process.

Here is an [article](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/) about this rule(in case someone doesn't know):

image

tyn1998:
> * I created a sample dataset only with CI data(google org data in 2015 and 2016), with `sample_data/sql_files/ci_test.sql` file.
>* Run a public ClickHouse instance for CI test and inject into npm test script.

I have always been wondering what is the best practice to test with SQL queries in CI. A fake database for development is good but is there a better way? I have no idea due to a lack of experience.

Hi @zhicheng-ning, do you have related experience?

package.json
```diff
@@ -14,7 +14,7 @@
     "notebook:win": "npm run build && docker pull docker-hub.x-lab.info/opendigger/open-digger-jupyter-nodejs:1.0 && docker run -it --rm -p 8888:8888 -v %cd%:/home/node/notebook docker-hub.x-lab.info/opendigger/open-digger-jupyter-nodejs:1.0",
     "pull-label-file-test": "tsc && node lib/ci/pull_label_file_test.js",
     "cron": "npm run build && node --max-old-space-size=8192 lib/cron/index.js",
-    "clean": "rm -rf dist"
+    "test": "CLICKHOUSE_HOST=http://106.15.39.211:8123 CLICKHOUSE_USERNAME=default mocha"
```
zhicheng-ning, line 17: It means that the test clickhouse server can be accessed by anyone?

zhicheng-ning:
> Hi @zhicheng-ning, do you have related experience?

I'm sorry that I just have a little unit test experiences.

frank-zsy:
> I have always been wondering what is the best practice to test with SQL queries in CI. A fake database for development is good but is there a better way? I have no idea due to a lack of experience.

Actually we can do tests with online server, but the workflow should not be triggered automatically by PR. Just like Kubernetes, a PR should be reviewed by reviewers first and then a committer can trigger an online test process manually, with this process, the triggered workflow can access the secrets and use online database server to test the SQLs.

And another solution is to add a database instance on test rather than a long term running instance, for integration test, the environment should not depend on other services, so fake services are common in tests.

But for OpenDigger right now, I think a sample database server is good enough for unit tests.

package.json
frank-zsy: Yes, if we want to test SQLs on PR and push automatically, the service needs to be public.

Although I think I should create a readonly account for CI test rather than use default account which has write permission.

tyn1998:
> I'm sorry that I just have a little unit test experiences.

That's OK, thank you for your repliy!

> Actually we can do tests with online server, but the workflow should not be triggered automatically by PR. Just like Kubernetes, a PR should be reviewed by reviewers first and then a committer can trigger an online test process manually, with this process, the triggered workflow can access the secrets and use online database server to test the SQLs.
>
> And another solution is to add a database instance on test rather than a long term running instance, for integration test, the environment should not depend on other services, so fake services are common in tests.

That information really helps, thank you!

package.json
zhicheng-ning: > Although I think I should create a readonly account for CI test rather than use default account which has write permission.

sounds great.

frank-zsy:
@zhicheng-ning Thanks for the notice, as the database is public to all, I have changed some server config for the instance, which includes:

- Change `default` account to readonly permission level, so there will be no account have write access to the database which means the database is totally readonly now.
- Add a query execution time limit to 5 seconds, which means time consuming query should not be in OpenDigger or any PRs, a query exceeds 5 seconds will be killed.
- Add a limitation that any query should not read more than 20 columns at a time.

The config is:

``` xml
<readonly>1</readonly>
<max_execution_time>5</max_execution_time>
<max_columns_to_read>20</max_columns_to_read>
```

frank-zsy:
Can you review and approve this one? @xgdyp

frank-zsy:
Since PR change with workflow can not be merged by GitHub Apps, I will merge this manually.

Issue #1128, opened 2023-01-08
[Bug] image can't output in jupyter notebook


PureNatural:
This is the first time I have used the function in opendigger to do data analysis.

I want to add open rank data analysis of  application_domain repository in China_open_source_report_2022.ipynb.

I finish it according to the existing code, but the image could not be output. 

The data has indeed been obtained.

After I executed the npm install and npm run build commands, I started jupyter notebook.

I'm not sure if there is a problem with my local environment.

<img width="1110" alt="image" src="https://user-images.githubusercontent.com/29674550/211204639-f0f60d57-378e-4c44-b58c-9289c53a3e6e.png">

frank-zsy:
@PureNatural Can you comment your code as text form so we can try it out, or you can just raise a PR with the change, that could be better.

PureNatural:
Ok, thanks for your reply. I will do it.

PureNatural:
We shuold use `npm run notebook` to start the notebook instead of using `jupyter notebook`.

Developers who use opendigger for the first time need to pay attention.

PureNatural:
This document describes the use of opendigger clearly. I didn't notice it before.
Thanks! @tyn1998

xgdyp:
I test this in vscode jupyter notebook plugin. It did not render the image. But if we use our jupyter image, it can.
I also test the ijs-plotly lib. It can render images successfully.
![image](https://user-images.githubusercontent.com/37795442/211531483-fbf6c7c3-e178-4aa1-b831-6735835e2cc6.png)
If we define a variable and plot, it can render, too.
![image](https://user-images.githubusercontent.com/37795442/211531924-4357eee0-1aa6-4447-ac87-6a783156cb09.png)
But if we plot directly after get data in one cell, it can't.
![image](https://user-images.githubusercontent.com/37795442/211532204-bdcb6fc5-bbf8-49c4-bbd1-c7b525829f7e.png)
I'm not sure if it caused by `promise` or jupyter version.

xgdyp:
Do you have any idea on this? @tyn1998 @frank-zsy

frank-zsy:
@xgdyp I did not use vscode notebook plugin at all, it may involve the behavior of the plugin. You should always use our notebook image to run OpenDigger.

frank-zsy:
Is there any update on this one? @xgdyp

xgdyp:
Hi, I test some methods before but there is no use. I think we can close it.

PR #1131, opened 2023-01-09
feat:add application domain repository data analysis

PureNatural:
1. capitalize the first letter of label names.
2. add application domain repository data analysis.

There is a [issue](https://github.com/X-lab2017/open-digger/issues/1128) with image output


frank-zsy:
There should be some other errors in your env, all works fine in my local env.
image
Did you use `npm run notebook` to start the notebook? I assume you did, but I can not find out what else may go wrong here.

frank-zsy:
What Node.js version you use? You can run `node -v` to check out, you should use Node.js version at least 14.17

frank-zsy:
I have updated the data in the notebook and rerun all the results to update the data to 2022.12 so they can be used in the report. @PureNatural @xgdyp

xgdyp:
Thanks！

PureNatural:
> There should be some other errors in your env, all works fine in my local env. image Did you use `npm run notebook` to start the notebook? I assume you did, but I can not find out what else may go wrong here.

I found the cause of the problem.

I use _jupyter notebook_ to start the notebook instead of using _npm run notebook_.

When I use _npm run notebook_ to start the notebook, the image can be output.

Much thanks for your help. @frank-zsy

Issue #1089, opened 2022-12-09
[proposal] GitHub Bot data requirement


bifenglin:
There are many auto activities in GitHub, such as the apps in [https://github.com/marketplace](https://github.com/marketplace), and some of them are easy to identify and can accord to the column `user_type,` which divided actors into `User` and `Bot.` However, some Bots also have a `User` type, meaning we have to identify them manually. And these actors are not real human beings and have a high probability produced activities by auto. So I propose the collection of data requires actors' labels about is a Bot or not.

Some of the examples:
- VSCodeTriageBot
- SnoozeThis
- LombiqBot
- renovate-bot
- codestar-github-bot-1
- rjpagano
- zhangsansana

frank-zsy:
@bifenglin Are you going to raise a PR for this?

bifenglin:
> @bifenglin Are you going to raise a PR for this?

yeah, I have found lots of GitHub bot accounts. But I haven't found any docs for how to create a new type of label data until you reply.

> Yes, I think so, you can add a bot folder and index.yml in it. Name should be GitHub Bot and type is Bot, add these accounts into data.github_users as array is fine.  

Thanks for this advice.

bifenglin:
And I have found some accounts belonging to a team or a company, such as `magento-engcom-team` and `fabricteam`. And I found these also a good dataset that can find some interesting findings, like lots of them both have auto activity and human activity. However, it also not appears in our label data.

frank-zsy:
closed by #1132

PR #1132, opened 2023-01-09
feat: add bot-lable data #1089

bifenglin:
1. create `bot` folder and  the template
2. add some bot-label data

### DIFF
--- a/labeled_data/bot/index.yml
+++ b/labeled_data/bot/index.yml
@@ -1,0 +1,11 @@
+name: GitHub Bot
+type: Bot
+data:
+  github_user:
+    - 105309205 # user:VSCodeTriageBot
+    - 96656532 # user:SnoozeThis
+    - 8517910 # user:LombiqBot
+    - 25180681 # user:renovate-bot
+    - 53108186 # user:codestar-github-bot-1
+    - 71103415 # user:rjpagano
+    - 74294145 # user:zhangsansana
--- a/src/label_data_utils.ts
+++ b/src/label_data_utils.ts
@@ -6,7 +6,7 @@ const labelInputDir = '../labeled_data';
 const labelInputPath = path.join(__dirname, labelInputDir);
 
 const supportedTypes = new Set<string>([
-  'Region', 'Company', 'Community', 'Project', 'Foundation', 'Tech-0', 'Tech-1', 'Tech-2', 'Domain-0'
+  'Region', 'Company', 'Community', 'Project', 'Foundation', 'Tech-0', 'Tech-1', 'Tech-2', 'Domain-0', 'Bot'
 ]);
 
 const supportedKey = new Set<string>([


frank-zsy:
@bifenglin You should check the existing label data, there are several errors in this PR.

- The extension of the file should be `yml` but not `yaml`, with the file you added, the file is not included in label data.
- The `data` field should not be simple array, you should add a sub filed `github_user`, just like [X-lab label data](https://github.com/X-lab2017/open-digger/blob/master/labeled_data/communities/xlab/index.yml#L11).
- In `github_user` field, the array element should be user id but not user login, you just parsed the ids in the issue, you should use the [output of the bot in the issue](https://github.com/X-lab2017/open-digger/issues/1089#issuecomment-1344019017).
- You should add a `Bot` type into the [label types](https://github.com/X-lab2017/open-digger/blob/master/src/label_data_utils.ts#L9), otherwise you will get an error in unit test.

bifenglin:
Thanks for advising, I have edited what ur said. And are there any problems?

labeled_data/bot/index.yml
```diff
@@ -0,0 +1,12 @@
+name: GitHub Bot
+type: Bot
+label:
+  - github_user
+github_user:
```
frank-zsy, line 5: The format is not correct, this line is redundant, and following lines need a more `tab`

labeled_data/bot/index.yml
```diff
@@ -0,0 +1,12 @@
...
+  - 105309205 # user:VSCodeTriageBot
+  - 96656532 # user:SnoozeThis
+  - 8517910 # user:LombiqBot
+  - 25180681 # user:renovate-bot
+  - 53108186 # user:codestar-github-bot-1
+  - 71103415 # user:rjpagano
+  - 74294145 # user:zhangsansana "
```
frank-zsy, line 12: tailing `"` should be removed.

labeled_data/bot/index.yml
```diff
@@ -0,0 +1,12 @@
+name: GitHub Bot
+type: Bot
+label:
```
frank-zsy, line 3: This should be `data` but not `label`, please check the existing data: https://github.com/X-lab2017/open-digger/blob/master/labeled_data/communities/xlab/index.yml

labeled_data/bot/index.yml
```diff
@@ -0,0 +1,13 @@
+name: GitHub Bot
+type: Bot
+data:
+  label:
+    - github_bot
```
frank-zsy, line 5: You should delete these 2 lines, there is no sub label called `github_bot`

bifenglin: So, do I needn't add a label? And the sub key uses `github_user` representing GitHub account is OK?

xgdyp:
And I think, in this label, we can add more bots whose name not like `%bot%`.

frank-zsy:
> And I think, in this label, we can add more bots whose name not like `%bot%`.

Yes, actually from OpenLeaderboard user ranking list, we can find lots of bot account which is not GitHub Apps on GitHub.

bifenglin:
That's no problem. And I have figured out plenty of bot accounts, And I will push pr in the future. So, this pr is good? I have fixed the label.

xgdyp:
I think it's no problem. @frank-zsy what do you think?
Otherwise, can you checkout bots in OpenLeaderboard first? I think this would improve OpenLeaderboard a lot！

xgdyp (approved):

PR #1152, opened 2023-01-22
fix:improve application domain label data

PureNatural:
I have checked more than 100 repositories to ensure that there are no obvious mistakes in the Open Rank top 20 repositories under each label.

When I checked the project, I found that the error rate of the repos under the non-software label was high. I will gradually revise more repos in the future.

When I try to rerun this [notebook](https://github.com/X-lab2017/open-digger/blob/master/notebook/China_open_source_report_2022.ipynb) by myself, I find that my local environment lacks data required for other cells on notebook.

Would you please rerun it again? @frank-zsy 

Much Thanks!

Happy lunar new year！


frank-zsy:
Thanks, I will rerun the notebook on this PR.

frank-zsy:
@PureNatural I have updated the notebook along with the time duration related metrics changes. Thanks for the data improvement and I think you should have the access to the online database service so you can perform any query in the notebook.

/approve

PR #1166, opened 2023-01-30
Remove unmaintained reports

tisonkun:
cc @frank-zsy 

If we have new reports, I'm glad to update the doc contents for them.


frank-zsy:
Thanks for the update, we will have more reports in the future with more visualization methods like:

- SVG images that can be directly embedded into documents.
- HTML report / dashboard about repo data.
- Hypercrx browser extension.

/approve

PR #1224, opened 2023-03-03
docs: Fix qrcode link error

sy-records:


frank-zsy:
Thanks for the fix

/approve

PR #1242, opened 2023-03-29
feat: add issue comment TFIDF similarity metrics and issue comment Ja…

bifenglin:
This PR adds the metric about actors' comment TFIDF similarity and actors' Jaccard similarity.

### DIFF
--- a/python/metrics/index.py
+++ b/python/metrics/index.py
@@ -2,6 +2,7 @@ import chaoss
 import activity_openrank
 import attention
 import related_users 
+import issue_comment_similartiy
 
 def getRepoActivity(config):
     return activity_openrank.getRepoActivity(config)
@@ -15,6 +16,10 @@ def getRelatedUsers(config):
     return related_users.getRelatedUsers(config)
 def getAttention(config):
     return attention.getAttention(config)
+def getIssueCommentTFIDFSimilarity(config):
+    return issue_comment_similartiy.getIssueCommentTFIDFSimilarity(config)
+def getIssueCommentJaccardSimilarity(config):
+    return issue_comment_similartiy.getIssueCommentJaccardSimilarity(config)
 
 # chaoss metrics
 def chaossCodeChangeCommits(config): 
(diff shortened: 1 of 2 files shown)


xgdyp:
hi @bifenglin , if this pr has not completed, you can convert it to draft PR

python/metrics/index.py
```diff
@@ -15,6 +16,10 @@ def getRelatedUsers(config):
     return related_users.getRelatedUsers(config)
 def getAttention(config):
     return attention.getAttention(config)
+def getIssueCommentTFIDFSimilarity(config):
```
xgdyp, line 19: It seems that `config` has never been used

python/metrics/issue_comment_similartiy.py
```diff
@@ -0,0 +1,141 @@
...
+from nltk.corpus import stopwords
+from sklearn.feature_extraction.text import CountVectorizer
+from scipy.linalg import norm
+from sklearn.feature_extraction.text import TfidfVectorizer
+
+def getSelectedActors(config):
+    """
+    TODO: get Selected Acotrs
+    """
+    sql = 'SELECT DISTINCT(actor_id) FROM opensource.gh_events')
```
xgdyp, line 15: I think we can add a limit clause such as `limit 10`

bifenglin: The function aims to find all users if have no selecting choice.

frank-zsy: `DISTINCT` selection from the whole table may lead to a OOM exception since the ClickHouse only has 128GB memory, we should definitely avoid this kind of SQL.

xgdyp:
hi, I think this pr can be converted into a case in  `notebook`. I'll rewrite it.

lhbvvvvv:
hi @bifenglin. I'm adding this to the `notebook` recently.  I think there is something to be confirmed below  :

- Both the code and the pr description are about the similarity of actors, so the pr title should be changed accordingly.
- `getRecent100Comment` in `getIssueCommentTFIDFSimilarity` is not defined .
- In `config`, do we want to pass the actors we want to query into the function?

PR #1288, opened 2023-05-09
fix: label group error

### LINKED ISSUE #1279: [question] the 'groupBy' parameter problem of the 'getRepoOpenrank' function
### Description

I can use the following method to get the top 10 openrank databases.
image

But when I add the `groupBy` parameter, the openrank from other domains will also be counted, such as big_data, cloud_native. 
image

If I want to get the openrank of each subdomain under the database domain, how should I choose the parameters here?

Thanks for your reply!

frank-zsy:
fix #1279 

`arrayJoin` function with 2 columns will cause a multiply of different values and result is wrong right now.

Use `arrayJoin... AS ` to fix the error and return corresponding label id and label name.

### DIFF
--- a/src/metrics/basic.ts
+++ b/src/metrics/basic.ts
@@ -243,7 +243,7 @@ export const getTimeRangeWhereClauseForClickhouse = (config: QueryConfig): strin
 }
 
 // clickhouse label group condition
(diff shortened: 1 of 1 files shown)


frank-zsy:
@xgdyp Can you give a look at this one?

Issue #1266, opened 2023-04-25
[OSPP2023] Data Quality Assurance & Metrics Explanation by case


xgdyp:
### Description

During the development of indicators, we need to verify whether the written sql gets the correct query results.Our previous approach was to select a repo and compare the data manually. At the same time, sometimes the data is missing, and we are often not sure whether our implements are wrong or the data is abnormal. This approach is inefficient and error-prone.

So we need some DQA mechanism to finish data-check and metric check automatically.

Further more, I think we can add a dataset and using this for metrics explanation.
e.g. the result of one metric runs on this dataset is 10.
and we can clearly understand which records make this metric 10, and display these records

references:
https://help.aliyun.com/document_detail/116897.html

frank-zsy:
As we already add a simple test framework for Node.js kernel in #1127 , I think we can expand the test to more detailed data.

The online test server contains data for whole `google` in 2015 and 2016, so we can use the data to generate a test static dataset for all the metrics:

- For single repo and for whole org
- For every month, quarter and year
- Group by different types

So we can use the static dataset for Node.js and Python kernel test for all kinds of queries with all kinds of configs, and the result should be identical.

yanchaomei:
Hi, I'm very interested in this project. Is there someone currently applying for the project or working on it? Also, would it be possible to contact via email or WeChat for further communication?

Zzzzzhuzhiwei:
I have relevant experience with Opendigger metrics, and I have a high interest in this project and hope to contribute my efforts. I really hope to get the chance to participate in this project on OSPP. ^ ^

dwxa:
I'm very interested in this project and I have relevant experience in data warehouse and data analysis and get a start on this project. Hope to have a chance to develop this project in OSPP.

frank-zsy:
@yanchaomei @Zzzzzhuzhiwei @dwxa Thanks for the interest in this issue, you can submit your application on the OSPP platform, if you have any questions about this task, just comment here and I will discuss with you guys about the details.

Lllllolita:
I am very interested in this project and hope to contribute my own efforts. I am researching data mining and data warehousing. I hope to have the opportunity to apply for this project in OSPP.

frank-zsy:
[...]

PR #1359, opened 2023-08-08
add the test data and script

Zzzzzhuzhiwei:
In this issue #1266, I get the data of  these seven metrics, and write the script to make sure the data is right.


test/data.test.js
```diff
@@ -0,0 +1,77 @@
+var openDigger = require('../src/open_digger')
+const fs = require('fs');
+const assert = require('assert');
+const _ = require('lodash');
+const _isEqual = require('lodash/isEqual');
+const FILENAME = '../test/testdata.json';
```
frank-zsy, line 6: For flexibility concern, I would like to suggest another file structure.

- Put all the data to `testdata` folder.
- Use `dataKey` as the file name of each data.
- In the file, we could provide a valid JSON object with the result and also the SQL to generate the result.

With above structure, we can add lots of test data in different files and easily to check if the data is right or not as well as the corresponding SQL.

Zzzzzhuzhiwei: I roughly understand, which folder is the `testdata` folder under?

Zzzzzhuzhiwei:
Now I put the `testdata` folder under the `test` folder and first generate the SQL and results of `issueClosed`.

frank-zsy:
Yes, this is exactly that I mean. And we definitely should design the naming rule here. I think we can all use underscore naming style. And all the parameters should be included in the file name to make sure no file data conflicts.

Zzzzzhuzhiwei:
Do the file names and parameters in the file both need to use underscore naming style，and does the file name need to contain the parameters: OrgID, limit, startyear, startmonth, endyear, endmonth and groupTimeRange?

frank-zsy:
> Do the file names and parameters in the file both need to use underscore naming style，and does the file name need to contain the parameters: OrgID, limit, startyear, startmonth, endyear, endmonth and groupTimeRange?

It depends on what we need to test. Org id is not necessary, year and month may not need to be included, just set to a constant is OK. But groupTimeRange is surely needed, and order, limit, limit type, group type(repo, org, label) are all needed.

Zzzzzhuzhiwei:
`issueClosed_desc_3_all_repo_month.json` ,from front to back are metric, order, limit, limitoption, groupby, groupTimeRange. Is this file naming format acceptable？

frank-zsy:
`issue_closed_...` is better, I think the metrics naming can be changed to underscore style too.

Zzzzzhuzhiwei:
I will temporarily set the `order` to `desc` and the `limitOption` to `all` to generate the metric results, in order to avoid uncertainty in the results.

frank-zsy:
That's OK, actually for some metrics with the problem, we can skip for now.

Zzzzzhuzhiwei:
I have added the testdata with determined results into files.

Zzzzzhuzhiwei:
Hi @frank-zsy , could you please review my PR? Your feedback would be greatly appreciated!

frank-zsy:
@Zzzzzhuzhiwei Sure, I will review this PR today.

frank-zsy:
@Zzzzzhuzhiwei Actually we are already now using `mocha` and `ts` extension to run the unit tests, so can you change the `js` file to `ts`, because right now your test file are not running in `npm test` command.

Zzzzzhuzhiwei:
Thank you for your suggestion, I will change the `js` file to `ts` today.

Zzzzzhuzhiwei:
hi @frank-zsy, I have changed the `js` file to `ts`, and my test file are running in `npm test` command. Thank you for your guidance very much!

frank-zsy:
@Zzzzzhuzhiwei The result looks good, I think you can delete the old `metrics.test.ts` and rename the new file to `metrics.test.ts`.

test/metrics.test.ts
```diff
@@ -1,210 +1,316 @@
...
-    // order works fine
-    if (options?.index !== undefined) {
-      assert.strictEqual(parseFloat(result[0][key][years - 1][options.index]) >= parseFloat(result[option.limit - 1][key][years - 1][options.index]), true);
-    } else {
-      assert.strictEqual(parseFloat(result[0][key][years - 1]) >= parseFloat(result[option.limit - 1][key][years - 1]), true);
+function deepEqual(a: any, b: any): boolean {
+  return _.isEqual(a, b);
+}
+
+async function validate_data(
```
frank-zsy, line 10: We should keep identical naming style, the function name should be `validateData`.

frank-zsy:
All looks good now, thanks for the contribution~

/approve

PR #1369, opened 2023-08-23
label: add more bot accounts

frank-zsy:
Add more bot accounts into label data. The result is from global OpenRank without first round labeling.


xgdyp:
/approve

Hi Frank, are these bots detected by graph algorithms? Or just filtered them manually?

frank-zsy:
@xgdyp Every time I labeled bot accounts, I will run global OpenRank again without the labeled bot accounts activity. And then I will fetch top 100 OpenRank user accounts for every month from 2015.1 to now. Then I will manually filter the results to find out automation accounts by names or their historical activities. It surely took a long time but the work is worthy, I think after few rounds, we will see real developers on OpenLeaderboard.

xgdyp:
Got it, thanks.

PR #1401, opened 2023-10-09
fix: fix the file path and throw error if the file not found

### LINKED ISSUE #1400: [Bug] The metrics test has some errors
### Current Behavior

The file path is not right so the test will fall into pass but with correct file path, some of the tests will fail.

### Expected Behavior

With right file path the test should all pass.

### Any Additional Comments?

_No response_

frank-zsy:
fix #1400

Fix the file path and throw error if test data file not found.

### DIFF
--- a/test/metrics.test.ts
+++ b/test/metrics.test.ts
@@ -1,7 +1,7 @@
 const openDigger = require('../src/metrics/index');
-import * as fs from "fs";
-const assert = require("assert");
-import * as _ from "lodash";
+import { existsSync, readFileSync } from 'fs';
+import assert from "assert";
+import _ from 'lodash';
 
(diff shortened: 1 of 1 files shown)


frank-zsy:
@Zzzzzhuzhiwei I did some minor fixes of the test file, and right now most of the tests pass but there are 29 tests fail. Can you give a look on this.

Zzzzzhuzhiwei:
It seems that this part of the data may have some problems, and I will correct it.

Zzzzzhuzhiwei:
hi @frank-zsy, if test data file are not found because the data is random or `Invalide group by label: null`

frank-zsy:
Ok, then we can remove the `throw error` part. After that how many tests fail?

PR #1403, opened 2023-10-09
fix: update testdata

### LINKED ISSUE #1400: [Bug] The metrics test has some errors

Zzzzzhuzhiwei:
fix some wrong testdata

fix #1400


Zzzzzhuzhiwei:
ok, I have modified the test script.

test/metrics.test.ts
```diff
@@ -12,29 +12,27 @@ async function validateData(
...
     if (!equal) {
       console.log(jsonData["modifiedOption"]);
       console.log(apiFn);
-      console.log("data1:", JSON.stringify(data1WithoutDetail));
-      console.log("data2:", JSON.stringify(data2WithoutDetail));
+      console.log("expected data: ", JSON.stringify(expectedDataWithoutDetail));
+      console.log("query result data: ", JSON.stringify(queryResultDataWithoutDetail));
     }
     assert(equal);
-  }
```
frank-zsy, line 37: For the missing files, should also has a log to indicate that the option is not supported.

frank-zsy:
/approve

Please add `fix` or `close` issue command in you PR body next time.

PR #1409, opened 2023-10-16
doc: add repo topic column

### LINKED ISSUE #1407: [Feature] Add topics columns into ClickHouse schema
### Description

We just ignored that actually from around 2021.11, the repo data in pull request related event logs started to contain the [`topics`](https://github.com/topics) of the repo. We should add the column into the database so we can analysis the topic related data without accessing the GitHub API anymore. This would be a huge help for analyzing global data and technology trends.

frank-zsy:
close #1407 

Add repository topics column into ClickHouse schema and all the data has been imported.


frank-zsy:
How to use:

image

Like shown in the image, we can use OpenRank weighted topics for a year and give a topic influence trend for all topics.

For 2022, the `hacktoberfest` topic is the most influential topic on GitHub and we can see a great increase for this topic in 2022.10 which makes sense since the event was held in Oct. every year.

image

PR #1423, opened 2023-11-04
feat: update user info for all users in github log

frank-zsy:
This PR will make the user update cron task to update information for all the users in GitHub log events but not only the exported ones.


frank-zsy:
The strategy for user info update process is:

- The cron task will run every 10 minutes for 2000 users.
- Firstly, we will update the users in export table. The total count is about 300 thousands. We will ensure to update their information once a month.
- When the export users update finished, we will update the information for those who have collaboration on GitHub like issue or PR events. The total is about 18 millions and we will ensure to update their information once a year.
- Lastly, we will update others users appeared in GitHub event log who may only star or fork repos on GitHub. The total count is about 59 millions. We will ensure to update their information at least once.

PR #1493, opened 2024-01-11
feat: add community openrank cron task

### LINKED ISSUE #1492: [Cron] The `community_openrank` process should also be part of OpenDigger
### Description

Right now, global_openrank script is part of OpenDigger to calculate the global OpenRank for all repos and users, we also should include the community_openrank script in OpenDigger which is now in a private repo of mine.

This will simplify the data process in every month and make the code easy to maintain.

### Cron Expression

0 0 14 1 * *

frank-zsy:
close #1492 

Add community_openrank task, since with 12 concurrent workers and 2GB memory for each worker, change the memory setting to 25.6GB for cron task command.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2303,6 +2303,11 @@
         "whatwg-url": "^5.0.0"
       }
     },
+    "node-worker-threads-pool": {
+      "version": "1.5.1",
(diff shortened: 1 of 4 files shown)


frank-zsy:
Though currently, we do not use `ReactionEvent` to add weight to issues and PRs right now, will be added later.

Issue #1478, opened 2024-01-09
OpenDigger Biweekly Meeting: 2024-01-10


Zzzzzhuzhiwei:
### Description

Meeting

🕣Time: 12:00~12:30, Wednesday of every two weeks
📞Tencent Meeting：567-2249-5032([link](https://meeting.tencent.com/dm/jmNIO1H667bc))
Agenda

The overview of last two weeks
In the last two weeks, we add lots of labeled data, including (#1466, #1467, #1472, #1475, #1476, #1477) ,prepare data for the annual report.
in the #1462, we add new metric: chaoss contributor into opendigger.
in the #1463, we want to find solution of implementing the python kernel. When updating Node.js kernel we need to manually migrate the changes to Python kernel.

PR #1498, opened 2024-01-14
Add pycjs, the python caller of js based on node_vm2

### LINKED ISSUE #1463: [Feature] Solution of implementing the Python kernel
### Description

Should we implement the entire python kernel function or only the python caller interface of the nodejs kernel?

birdflyi:
Fix #1463

# Purpose

- Add a more low maintenance costs and safer python kernel to replace the old kernel `python` and `python_v2`.

# Changes

- Add Func: Add the python caller of js 'pycjs' with the node_vm2 sandbox environment.
- Add a python kernel ipynb to call functions in the open_digger module as a simple example.

# Notes
- `pycjs_node_vm.py` is the core of pycjs. It can be used for other projects similar to the js2py type.
- The parameter kwargs of function `get_export_module` offer many display options for the variable `df_export_module`, which is useful when you want to know if a function is available or to lookup the call path of the functions.
- We need no migration of the changes to Python kernel as metioned in https://github.com/X-lab2017/open-digger/issues/1478#issue-2072446725 anymore：This kernel uses a virtual machine to run JavaScript functions, and only exposes function call handles in the sandbox.


pycjs/open_digger.py
```diff
@@ -0,0 +1,60 @@
...
+        'root': '../',
+        # 'mock': { }
+    }
+}
+
+js_import_open_digger = """
+var openDigger = require('../src/open_digger.js');
+"""
+
+if __name__ == '__main__':
```
frank-zsy, line 34: What does this mean? Is this an unit test of this file?

birdflyi: Yes, it is. This an unit test of this file.
The code segment under the `if __name__ == '__main__':` statement will be executed only when the `.py` file is executed as the unique entrance for main:0 thread of a python application, which means it cannot be executed when the module is initialized from an import statement.
In this typical case, it will be executed only when the `open_digger.py` is directly executed: 
1. Run `python ./pycjs/open_digger.py`,
2. Run `python -m open_digger` in the pycjs folder, because relative module names is not supported.

---
Related Introduction and Experiments: 
The file `__init__.py` has the same output as any typical `.py` file when it comes to the `if __name__ == '__main__':` statement.
Special situation: the `__main__.py` file in a module.
```python
print("__main__.py is executed.")
if __name__ == '__main__':
    print("__main__.py __name__ is __main__!")
```
If the code above exists in `pycjs/__main__. py` and a module init file `pycjs/__init__. py` is exist, there will be 3 ways to execute it:
[√] 1. Run `python ./pycjs/__main__.py`
[√] 2. Run `python ./pycjs`, it will choose the `__main.py` in `./pycjs` as the unique entrance.
[√] 3. Run `python -m pycjs` in the pycjs folder as relative module names is not supported.

However, it is not supported in this way: run `python -m __main__` in the pycjs folder. I still cannot understand the case as it report an error about module specification: `Error while finding module specification for '__main__' (ValueError: __main__.__spec__ is None)` .

frank-zsy:
All looks good, I will merge the PR first and we should encourage developers to use this new kernel lib to find out if it works fine as well.

birdflyi:
I completely agree~

PR #1506, opened 2024-01-17
Add New Functions and Add examples: An almost covering all existing functions version of pycjs kerel with some preliminary testing

birdflyi:
# Purpose
- Improve the solution for the issue #1463 
- Enhance the functionality of pycjs, covering all existing metric functions.

# Changes
- Add Python functions to replace functions that cannot be covered by sandboxes: 
  - Replace functions related to Plotly in the ijavascript-plotly module
  - Replace functions using global web elements like '$$'
  - Replace functions using functions as formal parameters, which are not JSON serializable types or cannot be executed by vm sandbox. 
- Add new examples in node_vm2_pycaller.ipynb. 
- Simplify call methods: compared with the function `get_export_module`, the function `get_export_module_open_digger` defined in `open_digger.py` has more suitable default settings for open-digger.
- Update dependencies: Update the dependency items in `requirements.txt`.

# Notes
- Exception report: The function `openDigger.quick.showAll` seems has bug with parameter `repoName`. I tried to query the activities and openranks of `'apache/spark'`, however, it returns the data of an other repository in the memory. Any more, the function can easily trigger an exception [TOO_LARGE_ARRAY_SIZE](https://github.com/ClickHouse/ClickHouse/issues/54053) if a query without any `label` constrants in options.
- Testing: Many functions in anual reports can be implemented by python, which provides more ways to solve problems as well as more diverse examples~


frank-zsy:
LGTM, thanks.

And `openDigger.quick` is a legacy API which is deprecated right now and should be removed from the lib, I will do this soon.

frank-zsy:
And like fixed in #1507 with the default config time value, right now for Jan. 2024, if you do not set the `endYear` and `endMonth` manually, it will cause a bug in the SQL which leads to the `TOO_LARGE_ARRAY_SIZE` problem, it should be fixed after above PR merged.

birdflyi:
> And like fixed in #1507 with the default config time value, right now for Jan. 2024, if you do not set the `endYear` and `endMonth` manually, it will cause a bug in the SQL which leads to the `TOO_LARGE_ARRAY_SIZE` problem, it should be fixed after above PR merged.

Great! This kind of Exception always confuses me with a hint `TOO_LARGE_ARRAY_SIZE`.

PR #1530, opened 2024-03-05
Add Greptime company label info

tisonkun:
Please help review @frank-zsy. Also, I wonder how this label data helps ...

cc @killme2008

### DIFF
--- a/labeled_data/companies/greptime/index.yml
+++ b/labeled_data/companies/greptime/index.yml
@@ -1,0 +1,9 @@
+name: Greptime
+type: Company
+data:
+  platforms:
+    - name: GitHub
+      type: Code Hosting
+      orgs:
+        - id: 103400677
+          name: GreptimeTeam


frank-zsy:
Thanks for this PR, LGTM.

If you add `Greptime` as a company label in OpenDigger, data generation for every month in [OpenLeaderboard](https://open-leaderboard.x-lab.info/) will include it as a company item. And if it is a Chinese company, actually you can also add it into [`regions/CN` label](https://github.com/X-lab2017/open-digger/blob/master/labeled_data/regions/CN.yml#L71), so it will also appear in the Chinese ranking list.

Further more, in all the open source report OpenDigger supports, we will also support to treat this org as a company.

tisonkun:
Thank you! Shall we merge it? CI failed but doesn't seem related to this issue. Perhaps we'd revisie the CI workflows later ..

frank-zsy:
Of course, merged.

PR #1541, opened 2024-03-28
fix: fix wrong name for change name repos.

### LINKED ISSUE #1540: [Bug] The data is not intact for some repositories that changed their names
### Current Behavior

As mentioned in https://github.com/hypertrons/hypertrons-crx/issues/776 , the activity.json data for https://github.com/LlamaFamily/Llama-Chinese is not intact.

### Expected Behavior

Intact data for any repo, especially for those which changed their names.

### Any Additional Comments?

Underlying SQL is not right for activity index and maybe other metrics, need to check.

frank-zsy:
fix #1540 

For all the metrics that has a inner query block in `getGroupIdClause`, we should use `max(created_at)` as the column to determine the `name` column rather than use `month` or other aggregated column.

### DIFF
--- a/src/metrics/chaoss.ts
+++ b/src/metrics/chaoss.ts
@@ -255,7 +255,7 @@ FROM
 (
   SELECT
     ${getGroupTimeClause(config, byCol)},
-    ${getGroupIdClause(config)},
+    ${getGroupIdClause(config, 'repo', 'last_active')},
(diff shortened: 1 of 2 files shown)


frank-zsy:
As the CI has broken, I have ran the test in local env with monthly export task, merge first and will fix if there are some minor errors in the future.

PR #1633, opened 2024-11-07
label: update antgroup labels

xiaoya-yaya:
I have updated the labels of opensource projects sponsored by Ant Group


frank-zsy:
Thanks for the update.

PR #1666, opened 2025-01-03
label: add selectdb company

QuakeWang:
add selectdb company


frank-zsy:
Thanks for the PR.

I am not quite familiar with Apache Doris, but I think the project is initiated by Baidu at first, although the core team found the SelectDB company later, but I think we should keep the project initiator as Baidu, WDYT?

frank-zsy:
Since the `doris` project label existed under `baidu`, I reuse the project label for `selectdb`.

